PR #47, opened 2019-10-23
[main.py] Debug multi-file inferencing.

ntadimeti:


main.py
```diff
@@ -404,48 +406,49 @@ def main():
...
-        _logger.info("Waiting for writer to finish...")
-        write_proc.join()
-        #############################################################
+        for x in range(0,len(files)):
+            infile = [files[x]]
+            if args.infer:
+                args.infer_files = infile
+                _logger.debug("Inference data: ", args.infer_files)
+            else:
+                args.val_files = files
```
tijyojwad: Should this be `infile` instead of `files`?

main.py
```diff
@@ -297,17 +297,19 @@ def eval_worker(gpu, ngpu_per_node, args, res_queue=None):
...
 
     # We need dimensions of the inference dataset.
     if args.infer:
         files_path = args.infer_files
     else:
         assert False, "writer called but infer = False. Not sure what file to write?"
+
     total_size = 0
     write_dimension = 0
     for fname in files_path:
```
tijyojwad: Since the writer is now for just one file at a time, I don't think we need this loop. It would only be writing the output for one of the input files, so that's all the info it needs

main.py
```diff
@@ -297,17 +297,19 @@ def eval_worker(gpu, ngpu_per_node, args, res_queue=None):
...
+        prefix = os.path.basename(fname).split(".")[0]
+        result_path = os.path.join(args.exp_dir, prefix + "_" + args.result_fname)
```
tijyojwad: I think the `prefix` is used to construct the `exp_dir` directory already?

main.py

main.py
ntadimeti: correct, done.

main.py
ntadimeti: I don't think it is. I double checked. It overwrites the same files multiple times if I don't add this prefix.

tijyojwad: Yes, I see now. I was confusing it with another `prefix` from a different portion of the code. This looks good.

main.py
```diff
@@ -404,48 +405,49 @@ def main():
...
-
-        # finish off writing
-        #############################################################
-        res_queue.put("done")
-        _logger.info("Waiting for writer to finish...")
-        write_proc.join()
-        #############################################################
+        for x in range(0,len(files)):
+            infile = [files[x]]
+            if args.infer:
```
tijyojwad: Since you brought up the `args` issue, I'm wondering if we can do a little bit of cleanup for it right now. Instead of updating the `infer_files` or `eval_files` argument, how about updating the signature of the `writer` function to take in the corresponding input file path directly in line 426? Then the writer function won't have to do a list index [0] query

main.py
```diff
@@ -296,24 +296,23 @@ def eval_worker(gpu, ngpu_per_node, args, res_queue=None):
              best_metric=best_metric, res_queue=res_queue, pad=args.pad)
 
 
-def writer(args, res_queue):
-    result_path = os.path.join(args.exp_dir, args.result_fname)
-    myprint("Writer process standing by... %s" % result_path, color='yellow')
+def writer(args, res_queue, files_path):
```
tijyojwad: another pedantic comment - instead of `files_path` (since it's only single file now), let's call it `input_file_path`?

tijyojwad:
@ntadimeti left a couple of styling comments. Approving the PR, so once those comments are resolved and branch is updated with master, feel free to merge!

PR #50, opened 2019-10-28
Compress bedGraph output before writing to file

avantikalal:
In postprocessing, each genomic interval is expanded to single-base resolution and a score is added to each base, creating a dataframe with one row per base, which is then written to bedGraph. 

In this change, bases with the same score are combined into a single row of the dataframe. This takes more processing time, but overall reduces write time to bedGraph and subsequent bedGraph>bigWig conversion time.


claragenomics/io/bedgraphio.py
```diff
@@ -18,34 +18,62 @@
...
+        expanded_df: Pandas dataframe containing chrom, start, end, score at base resolution.
+        positive (bool): if True, only regions with score>0 are retained
     Returns:
-        bg: pandas dataframe containing expanded intervals and scores where score>0.
+        intervals_df: Pandas dataframe with same columns; bases with same score are combined into one line.
     """
+    # For each base, attach the score assigned to the previous base
+    expanded_df['prevscore'] = [-1] + list(expanded_df['score'])[:-1]    
+    # Select bases where score changes - or the last base
+    intervals_df = expanded_df[(expanded_df['score'] != expanded_df['prevscore']) | (expanded_df.index==len(expanded_df)-1)].copy()
```
tijyojwad: Are these scores always integers or can they be floating point number? If they're always integer, then feel free to resolve this comment :).

If they can be floating point numbers, then it's best to use some proximity comparison - like that mentioned in https://docs.python.org/3/whatsnew/3.5.html#pep-485-a-function-for-testing-approximate-equality

ntadimeti:
@avantikalal  Do you have the run-time of the postprocess before and after this change ? 

Just trying to calibrate myself :)

claragenomics/io/bedgraphio.py
tijyojwad: ping @avantikalal

avantikalal: They will be integers in all of our use cases right now. It is theoretically possible that in some future use case they may not be integers, but as I don't know what that would be I think we can merge this.

avantikalal: In `postprocess.py` we either round or threshold the scores before applying this function, so small differences will be removed.

tijyojwad: Perfect, thanks @avantikalal !

tijyojwad:
@avantikalal - @ntadimeti ran your version of the code, and saw significant perf improvement in postprocess.py (on the larger dataset)

```
Before - 68m
After - 6m 40s
```

PR #51, opened 2019-11-01
WIP: [postprocess] Save inference results from main.py into bigwig format bypassing h5

ntadimeti:
1. Remove postprocess.py
2. Dump the inference results to bigwig directly from main.py
run.sh 
on master : 
this PR :


postprocess.py
```diff
@@ -92,51 +94,61 @@ def parse_args():
...
+def df_to_bigwig(batch_data, bwoutputfilename):
+    sizetuple = []
+    uniq_chroms = batch_data.chrom.unique()
+    with open(args.sizes_file, "r") as sizefile:
+        sizes  = sizefile.readlines()
+        for size in sizes:
+            size = size.strip().split("\t")
+            if size[0] in uniq_chroms:
+                sizetuple.append((size[0], int(size[1])))
+    bw = pyBigWig.open(bwoutputfilename, "w")
```
tijyojwad: this generate a .bw file per batch. there's also a `combiner` function that then combines all the intermediate files per batch into a single `.bw`. Does that work okay for multiple .bw files? Since each file has it's own header I'm wondering how well that's being handled in the combiner

postprocess.py
```diff
@@ -92,51 +94,61 @@ def parse_args():
...
+            # Sometimes np.around doesn't work with float32. To investigate.
+            scores = np.around(scores, decimals=args.round)
+
+        # if the batch contains values > 0, write them
+        if (scores > 0).any():
+            # Select intervals corresponding to batch
+            batch_intervals = intervals.iloc[start:end, :]
+            # Expand each interval and combine with scores
+            batch_bg = intervals_to_bg(batch_intervals, scores)
+            df_to_bigwig(batch_bg, outfilename)
```
tijyojwad: as discussed earlier, let's keep support for both `df_to_bigwig` and `df_to_bedgraph` depending on an input argument?

postprocess.py
ntadimeti: Good catch. It's doing a fine job as far as I can tell. I can open the file once and send in the object if that's a better coding practice.

postprocess.py
ntadimeti: Yes, I aim to do that as a separate PR if that's okay.

example/run.sh
```diff
@@ -136,83 +136,87 @@ echo ""
...
+    --round 3 \
+    --weights_path $out_dir/HSC.5M.model_latest/model_best.pth.tar \
+    --out_home $out_dir --label inference \
+    --result_fname HSC.5M.output.probs \
+    --model resnet --nblocks 5 --nfilt 15 --width 50 --dil 8 \
+    --nblocks_cla 2 --nfilt_cla 15 --width_cla 50 --dil_cla 10 \
+    --task both --num_workers 0 --infer_format bigiwg
 
-macs2 bdgpeakcall -i $out_dir/HSC.5M.output.probs.bedGraph -o $out_dir/HSC.5M.output.peaks.narrowPeak -c 0.5
+macs2 bdgpeakcall -i $out_dir/inference_latest/test_data_HSC.5M.output.probs.peaks.bedGraph -o $out_dir/inference_latest/test_data_HSC.5M.output.peaks.narrowPeak -c 0.5
```
tijyojwad: is this `macs2` call supposed to run the output of the previous infer call? That runs seems to be outputing the data in `bigwig` format, but the macs call is taking in a `bedGraph` file?

main.py
```diff
@@ -295,6 +299,42 @@ def eval_worker(gpu, ngpu_per_node, args, res_queue=None):
              world_size=args.world_size, distributed=args.distributed,
              best_metric=best_metric, res_queue=res_queue, pad=args.pad)
 
+def df_to_bigwig(intervals, args, batch_data, outputfile):
```
tijyojwad: let's move this function into `bigwigio` for now? And a task to write unit tests for it so we have don't end up breaking it

main.py
```diff
@@ -28,6 +28,8 @@
 import time
 import numpy as np
 from cmd_args import parse_args
+import pandas as pd
+import pyBigWig
```
tijyojwad: once `df_to_bigwig` is moved into its own module this import can be removed too from this file (into the bigwigio file)

example/run.sh
ntadimeti: Regardless of the format requested, bedGraph files are generated. If --infer_format = bigiwig then bedgraph files are then converted to bigwig. Right now, since we have a need for both formats, I am not deleting the bedGraph files generated if --infer_format = bigwig. Makes sense ?

main.py

calculate_baseline_metrics.py
```diff
@@ -205,7 +205,8 @@ def parse_args():
     # Load data
     _logger.info("Loading data for classification")
     x_peaks = read_data_file(args.test_file, 1, intervals)
-
+    #for idx in range(0, len(x_peaks)):
```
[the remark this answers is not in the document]
ntadimeti: done.

main.py

example/run.sh
tijyojwad: Yep, makes sense!

cmd_args.py
```diff
@@ -79,6 +79,14 @@ def parse_args():
...
                         help='list of data files in the form of [file1, file2, ...];'
                         'or a single path to a folder of files')
+    parser.add_argument('--intervals_file', type=str, default="",
+                        help='bed file containing the chr and interval values for inference')
+    parser.add_argument('--sizes_file', type=str, default="",
+                        help='bed file containing the chr and max size values')
+    parser.add_argument('--round', type=int,
+                        help='round the output to this value')
+    parser.add_argument('--infer_format', type=str, default="bedgraph",
+                        help='save the inference output to bigiwig or bedgraph.')
```
tijyojwad: let's update this to say `additionally save the inference output to bigwig` and rename the option to `gen_bigwig` for now, till we exclusively generate one of the other

ntadimeti: done.

PR #62, opened 2019-11-15
 [main.py] Add parallelization to writer

ntadimeti:
On large dataset:
Run time before this PR : 100 mins to generate bedgraph + 50 mins to generate bigwig
Run time after this PR : 28 mins to generate bedgraph + bigwig

What changed ? 
- Added parallelization to writer. (brought the bedGraph write output to 28mins) 
- Rounded peaks to 2 and track to 0. (decreased the number of rows to be written to the output, thus improving overall compute)
- Added option to threshold peaks.


main.py
```diff
@@ -345,31 +339,87 @@ def writer(args, res_queue, input_file_path):
...
+                pool_size = 0
+
+                if args.num_workers == -1: # spawn pool of processes
+                    num_cpus = mp.cpu_count()
+                    pool_size = min(num_jobs, num_cpus)
+                else: # user specified # of processes
+                    pool_size = args.num_workers
+
+                pool = mp.Pool(pool_size)
+                tempfiles = [os.path.join(temp_dir, str(channel), "{0:05}".format(num+batch_num)) for channel in channels for num in range(num_jobs)]
```
tijyojwad: few things here - 
1. let's separate the prefix and suffix of the file name with `_` (e.g. `tempdir0_num`)
2. when doing (num + batch_num), is that sum required to be unique? Asking coz job num 0 + batch num 1 == job num 1 + batch num 0
3. let's rename this var to `temp_files` to keep it similar to other vars

main.py
```diff
@@ -345,31 +339,87 @@ def writer(args, res_queue, input_file_path):
...
+                tmp_batch_ranges = [[i*args.batches_per_worker, (i+1)*args.batches_per_worker] for i in range(num_jobs)]
+                tmp_batch_ranges[-1][1] = batch.shape[0]
```
tijyojwad: can you add a comment here that this is to force the last interval to capture the remaining data?

main.py
```diff
@@ -345,31 +339,87 @@ def writer(args, res_queue, input_file_path):
...
+                tmp_batch_ranges = tmp_batch_ranges*len(channels)
+               
+                all_channels = [channel for channel in channels for i in range(num_jobs)]
+                map_args = list(zip(tmp_batch_ranges, [item]*len(tmp_batch_ranges), all_channels, [intervals]*len(tmp_batch_ranges),
```
tijyojwad: instead of doubling all the ranges by number of channels, it might be more readable if we call map_args once with channel = 0, and another time with channel = 1. This way it becomes a bit more readable and avoids copying of the intervals every loop

main.py
```diff
@@ -318,25 +322,15 @@ def get_batch_bg(args, item, channel, intervals):
             
         # Expand each interval, combine with scores, and contract to smaller intervals
         batch_bg = intervals_to_bg(batch_intervals)
-    return batch_bg
+        df_to_bedGraph(batch_bg, outfile)
 
 
-def writer(args, res_queue, input_file_path):
+def writer(args, res_queue, prefix):
```
tijyojwad: can you add doctoring here explaining the arguments? I know this function didn't have it from the beginning, but probably a good time to start now :)

main.py
ntadimeti: it's not a prefix, it's a subdir. `tempdir/0/(num+batch_num)`. Makes sense ?

main.py

main.py
ntadimeti: done

main.py
ntadimeti: done

main.py
```diff
@@ -345,31 +339,87 @@ def writer(args, res_queue, input_file_path):
...
-        outfiles = [open(os.path.join(out_base_path + ".peaks.bedGraph"), "w")]
+        outfiles = [os.path.join(out_base_path + ".peaks.bedGraph")]
     elif args.task == "both":
         channels = [0, 1]
-        outfiles = [open(os.path.join(out_base_path + ".track.bedGraph"), "w"),
-                    open(os.path.join(out_base_path + ".peaks.bedGraph"), "w")]
+        outfiles = [os.path.join(out_base_path + ".track.bedGraph"),
+                    os.path.join(out_base_path + ".peaks.bedGraph")]
+
+    # Temp dir used to save temo files during multiprocessing.
```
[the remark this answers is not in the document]
ntadimeti: done

main.py
```diff
@@ -29,7 +30,8 @@
 import numpy as np
 from cmd_args import parse_args
 import pandas as pd
-
+import math
```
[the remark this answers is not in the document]
ntadimeti: done. Double check if I missed some of them.

main.py
```diff
@@ -345,31 +362,101 @@ def writer(args, res_queue, input_file_path):
...
+                pool = mp.Pool(pool_size)
+                tmp_batch_ranges = [[i*args.batches_per_worker, (i+1)*args.batches_per_worker] for i in range(num_jobs)]
+                # Force the last interval to capture remaining data.
+                tmp_batch_ranges[-1][1] = batch.shape[0]
+                all_intervals = [intervals]*len(tmp_batch_ranges)
+                all_items = [item]*len(tmp_batch_ranges)
+                for channel in channels:
+                    tempfiles = [os.path.join(temp_dir, str(channel), "{0:05}".format(num+batch_num)) for num in range(num_jobs)]
+                    # No thresholding for tracks
+                    thresholding = args.infer_threshold
```
tijyojwad: maybe it's the github rendering, but where is this `thresholding` variable used?

main.py
tijyojwad: My bad, didn't see the `makedirs` call up top! All good.

And I'm guessing 2 wouldn't matter because in each loop all files are combined into the outfile anyway

tijyojwad: Can you fix the var name to be `temp_files`?

ntadimeti: Yes sorry, I only gave half answer. Yes, 2 wouldn't matter. I will change the file name.

main.py
ntadimeti: not used, good catch. deleted it.

main.py

tijyojwad:
@ntadimeti changes look good, just waiting for CI to pass and metrics to be validated after adding default rounding to tracks

main.py
```diff
@@ -345,31 +362,99 @@ def writer(args, res_queue, input_file_path):
...
+                                            temp_files, [rounding[channel]]*len(tmp_batch_ranges)))
+                    else:
+                        map_args = list(zip(tmp_batch_ranges, all_items,
+                                            [channel]*len(tmp_batch_ranges), all_intervals,
+                                            temp_files, [rounding[channel]]*len(tmp_batch_ranges),
+                                            [args.infer_threshold]*len(tmp_batch_ranges)))
+
+                    pool.starmap(save_to_bedgraph, map_args)
+
+                for channel in channels:
```
tijyojwad: Just noticed another thing - we can put this combiner within the same for loop above, so we wrote and combine for a single channel at a time

Issue #48, opened 2019-10-24
[main.py] Remove modification of args in the script.


ntadimeti:
`args` should always reflect command line arguments and should not be modified internally in the code. Similarly adding new `args` also isn't ideal.

ntadimeti:
Also change the following: 

pass args.infer_files as a string to `writer` function if only one file is being passed. Currently various modules expect a list, but it's unnecessary since we loop over several input files instead of appending them. #issue4

ntadimeti:
#94 addresses this.

ntadimeti:
@tijyojwad  closing this.

PR #68, opened 2019-11-25
[main.py] Part-1/2, Do not pass args command into functions

ntadimeti:
Pass explicit arguments into functions instead of passing `args`. All except infer_worker, eval_worker and train_worker are changed so that they accept explicit input arguments. 

The `*_workers` are not changed because the list of arguments they need is very big and we would need a data structure (may be a dict) to zip those arguments together. This will be addressed in another PR.

### DIFF
--- a/main.py
+++ b/main.py
@@ -70,35 +70,55 @@ def set_random_seed(seed):
         mpu.model_parallel_cuda_manual_seed(seed)
 
 
-def get_losses(args):
+def get_losses(task, mse_weight, pearson_weight, gpu):
+    """
+    Return loss function. 
+    Args:
+        task : Whether the task is regression or classification or both.
+        mse_weight : Mean squared error weight.
+        pearson_weight : Pearson weight.
+        gpu : Number of gpus.
+    Return:
+        loss_func : list of loss functions for each task.
+    """
 
-    if args.task == "regression":
+    if task == "regression":
         loss_func = MultiLoss(['mse', 'pearsonloss'], [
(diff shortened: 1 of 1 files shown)


tijyojwad:
These are some changes for #48

main.py
```diff
@@ -272,7 +298,10 @@ def infer_worker(gpu, ngpu_per_node, args, res_queue=None):
         dist.init_process_group(backend=args.dist_backend, init_method=args.dist_url,
                                 world_size=args.world_size, rank=args.rank)
 
-    model = build_model(args)
+    model = build_model(args.model, args.rank, args.interval_size, args.afunc,
```
tijyojwad: for these long argument functions, let's use named arguments to make the usage clearer (and avoid potential bugs).

e.g. `build_model(model=args.modek, rank=args.rank, interval_size=args.interval_size,...)`. Same for the `writer` function below as well

ntadimeti: we call the writer function using mp.Process. Not sure if I can send in args in the above format ... changed build_model.

tijyojwad: https://stackoverflow.com/questions/11546858/python-multiprocessing-keyword-arguments I believe describes how to pass keyword args to multi process calls. Can you check if this works in your scenario?

ntadimeti: changed it, can you check now ?

ntadimeti: would also appreciate if you can test it quickly to make sure nothing unexpected comes along the way

PR #79, opened 2019-12-08
Changes to metrics calculation

avantikalal:
-Changes to bigwigio.py:
1. Allow reading of bigWig files to any NumPy type

-Changes to calculate_baseline_metrics.py:
1.  Removed 'task = both' option since the script has to be run separately for regression or classification
2. Allow loading from bigWig/h5 to multiple dtypes
3. Set type to fp16 for classification probabilities and int8 for classification labels when only calculating AUC; this reduced whole-genome AUC calculation time from 3hrs to 1.5 hrs
4. Calculate threshold-based classification metrics only if --threshold is provided
5. Calculate AUC metrics if --auc flag is provided.
6. Called F1 score from metrics.py

-Changes to metrics.py:
1.  Rewrote spearman correlation. The scipy spearman function fails on genome-length arrays. 
2. Added F1 score to metrics.py

-Increased rounding values for probability in main.py. This is temporary until we make this a command-line argument.

- run.sh: removed duplicate commands for inference without threshold. fixed the call to `peaksummary.py` to take thresholded peak outputs instead of non-thresholded.

### DIFF
--- a/calculate_baseline_metrics.py
+++ b/calculate_baseline_metrics.py
@@ -31,7 +31,7 @@ import pandas as pd
 import os
 import logging
 
-from claragenomics.dl4atac.train.metrics import BCE, MSE, Recall, Precision, Specificity, CorrCoef, Accuracy, AUROC, AUPRC, SpearmanCorrCoef
+from claragenomics.dl4atac.train.metrics import BCE, MSE, Recall, Precision, Specificity, CorrCoef, Accuracy, AUROC, AUPRC, SpearmanCorrCoef, F1
 from claragenomics.io.bigwigio import extract_bigwig_intervals
 
 # Set up logging
@@ -236,15 +236,9 @@ else:
             calculate_class_nums(
                 x_peaks, t, message="Bases per class at threshold {}".format(t))
             metrics = calculate_metrics([Recall(t), Precision(
-                t), Specificity(t), Accuracy(t)], x_peaks, y_peaks)
+                t), Specificity(t), Accuracy(t), F1(t)], x_peaks, y_peaks)
             print("Classification metrics at threshold {}".format(t) +
                   " : " + " | ".join([str(metric) for metric in metrics]))
-
-            # Calculate F1 score from precision and recall
-            rec = metrics[0].get()
-            prec = metrics[1].get()
-            f1 = 2*rec*prec/(rec + prec)
-            print("F1 score at threshold {} : {:7.3f}".format(t, f1))
 
     # Calculate AUC
     if args.auc is not None:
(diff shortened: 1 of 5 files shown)


calculate_baseline_metrics.py
```diff
@@ -200,46 +203,49 @@ def parse_args():
...
+                x_peaks, t, message="Bases per class at threshold {}".format(t))
+            metrics = calculate_metrics([Recall(t), Precision(
+                t), Specificity(t), Accuracy(t)], x_peaks, y_peaks)
+            print("Classification metrics at threshold {}".format(t) +
+                  " : " + " | ".join([str(metric) for metric in metrics]))
+
+            # Calculate F1 score from precision and recall
+            rec = metrics[0].get()
+            prec = metrics[1].get()
+            f1 = 2*rec*prec/(rec + prec)
```
tijyojwad: can you add F1 to metrics.py as well?

claragenomics/dl4atac/train/metrics.py
```diff
@@ -255,9 +257,14 @@ def __init__(self):
...
         x = Metric.convert_to_numpy(x)
         y = Metric.convert_to_numpy(y)
-        x = x.flatten()
-        y = y.flatten()
-        self.val = scipy.stats.spearmanr(y, x)[0]
+        x = np.apply_along_axis(rankdata,0,x)
+        y = np.apply_along_axis(rankdata,0,y)
+        xm = x.mean()
+        ym = y.mean()
+        np.subtract(x, xm, out=x)
```
tijyojwad: is this the in-place calculation that caused the issue with backprop during training?

avantikalal: No, the problem was in `CorrCoef` which is now changed back.

calculate_baseline_metrics.py
avantikalal: done

claragenomics/dl4atac/train/metrics.py
```diff
@@ -86,11 +87,14 @@ def __init__(self):
...
-        x_norm = x - torch.mean(x)
-        y_norm = y - torch.mean(y)
-
-        self.val = torch.sum(x_norm * y_norm) / (torch.sqrt(torch.sum(x_norm**2))
-                                             * torch.sqrt(torch.sum(y_norm**2)))
+        xm = x.mean()
+        ym = y.mean()
+        x = x - xm
+        y = y - ym
+        #x.add_(-1*xm)
```
tijyojwad: let's remove the commented lines of code. feel free to resolve this once the change is made

PR #80, opened 2019-12-09
[main] check dataset and interval size during inference

tijyojwad:


main.py
```diff
@@ -250,6 +250,12 @@ def infer_worker(gpu, ngpu_per_node, args, res_queue=None):
     infer_dataset = DatasetInfer(args.infer_files)
     infer_sampler = None
 
+    # Check that the intervals file and infer data have same number of entries.
+    intervals = pd.read_csv(args.intervals_file, sep='\t', header=None, names=['chrom', 'start', 'end'],
```
avantikalal: Why do we need to distribute this check to every GPU? Would it be better placed in `main()`?

tijyojwad: Good point. Moved it over

main.py
```diff
@@ -529,6 +529,15 @@ def main():
                 args.interval_size = f['data'].shape[1]
                 args.batch_size = 1
 
+            # Ensure the sizes for intervals file and infer dataset are same.
+            infer_dataset = DatasetInfer(args.infer_files)
```
avantikalal: This should only be run if `args.infer` is true, not `args.eval`.

main.py
tijyojwad: done

main.py
tijyojwad: done

PR #82, opened 2019-12-16
[main.py] Add a command line argument for rounding

ntadimeti:
Create a command line argument for main.py.


example/run.sh
```diff
@@ -141,7 +141,7 @@ python $root_dir/main.py --infer \
     --infer_threshold 0.5 \
     --weights_path $out_dir/HSC.5M.model_latest/model_best.pth.tar \
     --out_home $out_dir --label inference \
-    --result_fname HSC.5M.output \
+    --result_fname HSC.5M.output --rounding 0 3 \
```
tijyojwad: hmm this might be prone to user error, with the positions being flipped. I would propose breaking this into two options (adding to our existing list of a million options). So let's do this now. Then once we have the config file (with a base default config) users hopefully won't have too much trouble locating and updating these options.

ntadimeti: okay, makes sense.

ntadimeti: done.

PR #84, opened 2019-12-19
[main.py] Provide command line option for choosing best metric

ntadimeti:
Addresses issue 55 : https://github.com/clara-genomics/AtacWorks/issues/55

### DIFF
--- a/main.py
+++ b/main.py
@@ -35,6 +35,7 @@ from claragenomics.dl4atac.train.dataset import DatasetTrain, DatasetInfer
 from claragenomics.dl4atac.train.evaluate import evaluate
 from claragenomics.dl4atac.train.infer import infer
 from claragenomics.dl4atac.train.losses import MultiLoss
+from claragenomics.dl4atac.train import metrics
 from claragenomics.dl4atac.train.metrics import BCE, MSE, Recall, Specificity, CorrCoef, AUROC
 from claragenomics.dl4atac.train.models import *
 from claragenomics.dl4atac.train.train import train
@@ -96,12 +97,13 @@ def get_losses(task, mse_weight, pearson_weight, gpu, poisson_weight):
     return loss_func
 
 
-def get_metrics(task, threshold):
+def get_metrics(task, threshold, best_metric_choice):
     """
     Get metrics. 
     Args:
         task : Whether the task is regression or classification or both.
         threshold : the threshold for classification.
+        best_metric_choice : which metric to use for best metric.
     Return: #TODO
         metrics_reg :
         metrics_cla :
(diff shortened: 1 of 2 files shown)


main.py
```diff
@@ -420,7 +442,7 @@ def writer(infer, intervals_file, exp_dir, result_fname,
     channels = []
     outputfiles = []
     out_base_path = os.path.join(exp_dir, prefix + "_" + result_fname)
-    if task == "regression":
+    if task == "regression":i
```
avantikalal: Delete the "i"?

ntadimeti:
@tijyojwad  I don't think I can use the `getattr` method here to query for the existence of certain metric. This is because each metric is defined as a separate class. Whereas in losses, all losses are defined within a single class "MultiLoss". At this stage I think elif loops are most straight forward. WDYT ?

tijyojwad:
> @tijyojwad I don't think I can use the `getattr` method here to query for the existence of certain metric. This is because each metric is defined as a separate class. Whereas in losses, all losses are defined within a single class "MultiLoss". At this stage I think elif loops are most straight forward. WDYT ?

@ntadimeti I think this works - 
```
>>> from claragenomics.dl4atac.train import metrics
>>> MSE_class = getattr(metrics, "MSE")
>>> type(MSE_class)
<class 'type'>
>>> MSE_class
<class 'claragenomics.dl4atac.train.metrics.MSE'>
>>> RANDOM_class = getattr(metrics, "FOOBAR")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'claragenomics.dl4atac.train.metrics' has no attribute 'FOOBAR'
>>>
```

main.py
```diff
@@ -113,17 +115,17 @@ def get_metrics(task, threshold):
...
                        Specificity(threshold), AUROC()]
-        best_metric = metrics_cla[-1]
     elif task == 'both':
         metrics_reg = [MSE(), CorrCoef()]
         metrics_cla = [BCE(), Recall(threshold),
                        Specificity(threshold), AUROC()]
-        best_metric = metrics_cla[-1]
-
+    try:
+        best_metric = getattr(metrics, best_metric_choice)()
```
tijyojwad: The old code was using the reference to an existing metric object as the best metric, whereas this new code is instantiating a new metric object. I think this may cause some issues downstream because the best metric variable is likely not updated in the same way the other metrics are. I think the safest thing here would be to look for an instance of the best metric class in the metrics arrays and use that reference

ntadimeti: makes sense. made the changes. good catch

PR #89, opened 2020-01-08
Add config params config file

ntadimeti:
Add config file for convenient usage of different command line options. Change is backward compatible.
config file saved in experiment directory.

### DIFF
--- a/configs/config_params.yaml
+++ b/configs/config_params.yaml
@@ -1,0 +1,59 @@
+#
+# Copyright (c) 2020, NVIDIA CORPORATION.  All rights reserved.
+#
+# NVIDIA CORPORATION and its licensors retain all intellectual property
+# and proprietary rights in and to this software, related documentation
+# and any modifications thereto.  Any use, reproduction, disclosure or
+# distribution of this software and related documentation without an express
+# license agreement from NVIDIA CORPORATION is strictly prohibited.
+#
+
+
+#Learning args
+clip_grad: 0 
+lr: 0.0001 
+epochs: 5
+afunc: relu 
+mse_weight: 0.001 
+pearson_weight: 1
+poisson_weight: 0
+# Experiment args
+label: 'AtacWorks'
+out_home: './Cache'
+train: False
+infer: False
+resume: False
+eval: False
+infer_files: "None"
+intervals_file: "None"
+sizes_file: "None"
+infer_threshold: "None"
+reg_rounding: 0
+cla_rounding: 3
+batches_per_worker: 16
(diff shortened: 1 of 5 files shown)


configs/config_params.yaml
```diff
@@ -0,0 +1,64 @@
...
+#
+# NVIDIA CORPORATION and its licensors retain all intellectual property
+# and proprietary rights in and to this software, related documentation
+# and any modifications thereto.  Any use, reproduction, disclosure or
+# distribution of this software and related documentation without an express
+# license agreement from NVIDIA CORPORATION is strictly prohibited.
+#
+
+
+##USER NOTE: Certain fields are commented out as they are
```
tijyojwad: this user note is not needed anymore, right?

configs/config_params.yaml
```diff
@@ -0,0 +1,64 @@
...
+# Training args
+task: 'both'
+train_files: ""
+print_freq: 10
+bs: 8
+num_workers: 4
+checkpoint_fname: ""
+save_freq: 1
+# Dataset args
+pad:
```
tijyojwad: I'm surprised leaving it empty works instead of requiring a "None" string here. I think it would be better to instead pass "None" as a string in cases where we explicitly want None

configs/config_params.yaml
```diff
@@ -0,0 +1,64 @@
...
+transform: 'none'
```
tijyojwad: let's update this transform option to be 'None' as well so we have None standardized in our API (instead of one that's lower case only and one that's camel case)

or alternatively we can update the type_or_none_fn to compare against `str(val).lower()` so it's case insensitive

configs/config_params.yaml
ntadimeti: done.

configs/config_params.yaml
ntadimeti: done.

configs/config_params.yaml
ntadimeti: done.

tijyojwad: One last request - looks like the "None" string isn't really checked in the transform logic (only the log option for now). Let's just take out 'None' from the choices for transform  (or add update 'none' to be 'None')

ntadimeti: So, in the case of transform.. it expects a string "none". Not the python NoneType object. My bad I didn't notice it before. It should be left as "none" so that it's not converted to NoneType by our function.

ntadimeti: done. removed "none" as a choice.

PR #93, opened 2020-01-10
[pip] fix numpy issue during requirements installation

tijyojwad:


README.md
```diff
@@ -65,9 +65,12 @@ git clone --recursive https://github.com/clara-genomics/AtacWorks.git
 * Install pip dependencies
 
     ```
-    pip install -r requirements-pip.txt
+    pip install -r requirements-base.txt && pip install -r requirements-macs2.txt
     ```
 
+Note: The above non-standard installation is necessary to ensure the requirements for macs2 are installed
+before macs2 itself.
```
ntadimeti: Is it not working if the macs2 requirements are installed followed by other ones in the same file ?

tijyojwad: Unfortunately not. That's in fact how it was setup previously. But it doesn't work for the following reasons - 

1. when pip installs modules from a file, it doesn't actually `install` them. It first builds them all, and then installs them after the builds are done. https://stackoverflow.com/questions/5394356/how-to-specify-install-order-for-python-pip
2. The setup.py script for `macs2` (https://github.com/taoliu/MACS/blob/master/setup.py) calls `import numpy`, but since pip doesn't actually install numpy (from 1), this call fails.

The only way to make step 2 work is to first install numpy, and then install macs2. That's why we had to workaround by manually installing numpy first. But Breaking apart macs2 from the rest of the dependencies seems neater because macs2 is the offending module and not numpy.

ntadimeti: aah, makes sense!

PR #94, opened 2020-01-10
Clean up main.py

ntadimeti:
Move worker functions into their own module. 
Remove modification of args object.

### DIFF
--- a/worker.py
+++ b/worker.py
@@ -1,0 +1,269 @@
+#!/usr/bin/env python
+
+#
+# Copyright (c) 2020, NVIDIA CORPORATION.  All rights reserved.
+#
+# NVIDIA CORPORATION and its licensors retain all intellectual property
+# and proprietary rights in and to this software, related documentation
+# and any modifications thereto.  Any use, reproduction, disclosure or
+# distribution of this software and related documentation without an express
+# license agreement from NVIDIA CORPORATION is strictly prohibited.
+#
+
+# system imports
+import logging
+import os
+import random
+import sys
+import tempfile
+import time
+
+import h5py
+import multiprocessing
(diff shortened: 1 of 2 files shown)


worker.py
```diff
@@ -0,0 +1,274 @@
...
+    '%(levelname)s:%(asctime)s:%(name)s] %(message)s')
+_logger = logging.getLogger('AtacWorks-main')
+_handler = logging.StreamHandler()
+_handler.setLevel(logging.INFO)
+_handler.setFormatter(log_formatter)
+_logger.setLevel(logging.INFO)
+_logger.addHandler(_handler)
+
+
+def get_losses(task, mse_weight, pearson_weight, gpu, poisson_weight):
```
tijyojwad: for me the `get_metrics`, `get_losses` and `get_model` functions don't quite fit under a `workers` module, since those functions themselves are used by workers but not specific to workers. how about we leave those functions in main, and instead pass the losses/metrics/model objects into the worker functions as args?

ntadimeti: but they are specific to workers. They seem out of place in main.py as well. Consider these as helper functions ?

tijyojwad: Yeah, I think it's okay to consider them as helper functions. But since they've been moved to other modules, it would be good to make the following changes - 
1. remove the imports from main that aren't required anymore
2. update the API for the workers to take specific arguments instead of the `args` struct. since they're in a separate module now it's better to have fixed APIs for them

ntadimeti: keep in mind here, the args struct is the config object. So, I don't see anything wrong about passing the config object around. I removed code that modifies the args object.

ntadimeti:
rerun tests

PR #99, opened 2020-01-15
Interval reader

avantikalal:
Created functions to read interval files and chromosome sizes files, so that we load these files uniformly in all scripts.


tijyojwad:
FYI @ntadimeti

ntadimeti:
awesome ! thanks for doing this @avantikalal

PR #103, opened 2020-01-21
Motif model

### LINKED ISSUE #13: Save data in named Numpy fields
- name Numpy fields as noisy_data, clean_data, clean_peaks
- saving noisy data is not required once dynamic downsampling is merged.
- edit all data loading accordingly
- edit calculate_baseline_metrics accordingly

avantikalal:
**Extra inputs**
AtacWorks can now take as input any number of additional bigWig files ('layers') aside from the noisy ATAC-seq data. All of these will be concatenated as additional channels in the model input ('input'). The number of input tracks is supplied as an argument ('in_channels') to `main.py`.

h5 files now contain multiple datasets ('input' for inputs, 'label_reg' and 'label_cla' for labels) instead of combining everything into one dataset called 'data'.

Closes #13 .

**Always using --nolabel for inference**

I deleted some lines in DatasetInfer (dataset.py) because I don't understand their function. As far as I can see we never need DatasetInfer to supply labels - only the input.

Deleted unused DatasetEval.

run.sh included two commands to encode inference data, with and without labels. Since we never need labels to be encoded in the .h5 file for inference, I removed the command for encoding with labels.


bw2h5.py
```diff
@@ -116,61 +119,68 @@ def parse_args():
...
+# Write batches to hdf5 file
+_logger.info('Extracting data for each batch and writing to h5 file')
 with h5py.File(filename, 'w') as f:
     # Create a single dataset -- expand locations as we go along
     for i in range(batches_per_epoch):
+        batch_name = 'batch' + str(i)
         if i % 10 == 0:
-            _logger.info("batch: " + str(i) + " of " + str(batches_per_epoch))
+            _logger.info(batch_name + " of " + str(batches_per_epoch))
+            batch_data = {}
```
ntadimeti: is this supposed to be initialized inside the `if` block ?

bw2h5.py
```diff
@@ -116,61 +119,68 @@ def parse_args():
...
 
         # Subset intervals
         batch_intervals = intervals.iloc[batch_starts[i]:batch_ends[i], :]
 
         # Read noisy data
-        batch_data = extract_bigwig_intervals(
+        batch_data['x'] = extract_bigwig_intervals(
```
ntadimeti: key `x` sounds arbitrary. May be better to use a more defining keyword like `input`. Improves readability of code.

bw2h5.py
```diff
@@ -116,61 +119,68 @@ def parse_args():
...
+                layer_data = extract_bigwig_intervals(
+                    batch_intervals, layer_file, pad=args.pad
+                )
+                batch_data['x'] = np.dstack((batch_data['x'], layer_data))
+
         if not args.nolabel:
 
             # Read clean data: regression labels
-            clean_data = extract_bigwig_intervals(
+            batch_data['y_reg'] = extract_bigwig_intervals(
```
ntadimeti: what does `y` signify in `y_reg` ? May be better to have a more defining prefix.

bw2h5.py
```diff
@@ -116,61 +119,68 @@ def parse_args():
...
                 batch_intervals, args.cleanbw, pad=args.pad
             )
 
             # Read clean data: classification labels
-            clean_peak_data = extract_bigwig_intervals(
+            batch_data['y_cla'] = extract_bigwig_intervals(
```
ntadimeti: Let's replace `y` with something more descriptive.

claragenomics/dl4atac/dataset.py
```diff
@@ -121,12 +122,13 @@ def _get_generator(self):
                 sys.stdout.flush()
             rec = self.fh_data[file_id][local_idx - self.fh_indices[file_id][0]]
             sys.stdout.flush()
-            if len(rec.shape) == 1:
+            idx = yield {'idx':idx, 'x':rec}
+            #if len(rec.shape) == 1:
```
ntadimeti: can we delete the commented sections ?

claragenomics/dl4atac/dataset.py
```diff
@@ -121,12 +122,13 @@ def _get_generator(self):
...
-                idx = yield {'idx':idx, 'x':rec}
-            else:
+            #    idx = yield {'idx':idx, 'x':rec}
+            #else:
                 # Return 4 items -- IDX (for saving/tracing), input data, upsampled data, peaks/classifications
-                idx = yield {'idx':idx, 'x':rec[:,0], 'y_reg':rec[:,1], 'y_cla':rec[:,2]}
+                #idx = yield {'idx':idx, 'x':rec[:,0], 'y_reg':rec[:,1], 'y_cla':rec[:,2]}
 
 """
 # Is this even used?
```
ntadimeti: should we delete this whole part ? Is it functional at all ?

ntadimeti:
Additional things we discussed to be added to this PR : 
1. Add  --layersbw  usecase to example/run.sh
2. Updating example 7b, and calculate_baseline_metrics.py in example/run.sh
3. Remove test_data.h5 and use no_label.h5 for inference and bigwig files otherwise. 

Additionally, once PR#97 is merged, please run
--> flake8 --ignore=E901 <Atacworks source dir path>
--> pydocstyle --convention=google <Atacworks source dir path>

to ensure linting and documentation are correctly formatted.

claragenomics/dl4atac/dataset.py
avantikalal: I don't think so, but I'd like @tijyojwad to check and confirm.

bw2h5.py
avantikalal: fixed

ntadimeti: I meant the `batch_data` dictionary, sorry. Shouldn't it be defined outside the `if i % 10 == 0:` block ?

ntadimeti: just saw that you fixed it in other commit. Thanks

bw2h5.py
avantikalal: changed to 'input'

bw2h5.py
avantikalal: changed to 'label_reg'

bw2h5.py
avantikalal: changed to 'label_cla'

claragenomics/dl4atac/models/model_utils.py
```diff
@@ -124,16 +126,16 @@ def build_model(rank, interval_size, resume,
                               afunc=model_args.afunc, bn=model_args.bn)
     elif model_args.model == 'resnet':
         model = DenoisingResNet(interval_size=interval_size,
-                                afunc=model_args.afunc,
-                                bn=model_args.bn,
+                                afunc=model_args.afunc, bn=model_args.bn,
```
tijyojwad: let's move this back to be on two line. more readable that way

claragenomics/dl4atac/dataset.py
avantikalal: @tijyojwad do these commented lines serve any purpose?

bw2h5.py
```diff
@@ -126,65 +130,74 @@ def parse_args():
...
         # Read noisy data
-        batch_data = extract_bigwig_intervals(
+        batch_data['input'] = extract_bigwig_intervals(
             batch_intervals, args.noisybw, pad=args.pad)
 
+        # Add other input layers
+        if args.layersbw is not None:
+            # Read additional layers
+            layer_files = gather_files_from_cmdline(args.layersbw, ".bw")
+            for layer_file in layer_files:
```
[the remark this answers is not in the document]
avantikalal: Yes, it should be the same order in which the files are passed in. Or, if a path to a directory is given, it should be the order of files in the directory.

claragenomics/dl4atac/models/model_utils.py
avantikalal: Done.

bw2h5.py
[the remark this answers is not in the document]
avantikalal: Added help message in `bw2h5.py`

claragenomics/dl4atac/dataset.py
avantikalal: I deleted the commented lines. Have not deleted the DatasetEval section.

claragenomics/dl4atac/dataset.py
avantikalal: I deleted the commented lines. Have not deleted the DatasetEval section.

ntadimeti: I think it's safe to delete the DatasetEval as well. It's not being used anywhere in the code currently. @tijyojwad to confirm if I'm missing something.

claragenomics/dl4atac/dataset.py
tijyojwad: Let's delete dataset eval.as well

claragenomics/dl4atac/dataset.py
tijyojwad: Yeah I think okay to delete dataset eval

bw2h5.py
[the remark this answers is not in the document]
tijyojwad: Order of file in the directory is a bit ambiguous though. I think it's a good idea to enforce strict ordering from the user when passing the layers.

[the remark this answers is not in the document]
avantikalal: How would you do that? Sounds like this may need to be a separate PR?

claragenomics/dl4atac/dataset.py
avantikalal: deleted.

claragenomics/dl4atac/dataset.py
avantikalal: deleted.

bw2h5.py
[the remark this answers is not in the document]
tijyojwad: One way we can do it is name each of the layers individually when creating them. Then when training the model specify the ordering of the input based on the names of the layers. e.g. when creating the dataset we do something like `--layers low_cov:path_to_file_a quality_score:path_to_file_b ...`, and this will create separate keys in the h5 file called `low_cov` and `quality_score`. Now during training we can specify in model params config which of these layers you want to train with. This way we can generate a big h5 with all layers ahead of time, and train multiple models with different layers easily (for ablation studies for instance).

I would say we should do it in this PR since the API for the current implementation will be different than the one I'm suggesting. I'll let @ntadimeti take a call on this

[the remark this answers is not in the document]
ntadimeti: I agree, we should do it in this PR. Great idea @tijyojwad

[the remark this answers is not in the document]
avantikalal: @tijyojwad added naming of layers.

example/run.sh
```diff
@@ -210,15 +208,74 @@ echo ""
...
+    --prefix $out_dir/test_data_layers \
+    --nolabel
+
+#Train and validate model..."
+python $root_dir/main.py --train \
+    --train_files $out_dir/train_data_layers.h5 \
+    --val_files $out_dir/val_data_layers.h5 \
+    --out_home $out_dir --label HSC.5M.model.layers \
+    --layers ctcf_for \
+    --in_channels 2 \
```
tijyojwad: @avantikalal can the channels be automatically determined by how many layers need to be used?

bw2h5.py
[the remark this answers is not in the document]
tijyojwad: Awesome, changes look really good! Thanks @avantikalal

example/run.sh
avantikalal: @tijyojwad I suppose so. But I thought that the number of input channels should be included in the model parameters file as it's an important part of the model structure.

tijyojwad: Fair enough, I'm okay with it being there. Also adds another layer of sanity.

PR #109, opened 2020-01-31
Add end-to-end tests

ntadimeti:
This PR basically takes the example/run.sh and compares each step's output with an expected golden figure. 

We figured out through this that our models are not exactly reproducible. In another PR, I will work on that and then compare the Step 6 (inference) results. Currently, it's not feasible to test because the bedgraph files are not identical after inference.


tests/end-to-end/run.sh
```diff
@@ -0,0 +1,263 @@
...
+
+echo ""
+echo "Step 7b: Calculate metrics for peak classification after inference..."
+echo ""
+python $root_dir/calculate_baseline_metrics.py \
+    --label_file $out_dir/clean.peaks.bw \
+    --task classification \
+    --test_file $out_dir/inference_latest/no_label_infer_results.h5.peaks.bw \
+    --intervals $out_dir/result.holdout_intervals.bed \
+    --sizes $ref_dir/hg19.auto.sizes \
```
avantikalal: --sizes does not need to be provided for this command.

tests/end-to-end/run.sh
```diff
@@ -0,0 +1,263 @@
...
+
+echo ""
+echo "Step 7a: Calculate metrics for track coverage after inference..."
+echo ""
+python $root_dir/calculate_baseline_metrics.py \
+    --label_file $data_dir/HSC.80M.chr123.10mb.coverage.bw \
+    --task regression \
+    --test_file $out_dir/inference_latest/no_label_infer_results.h5.track.bw \
+    --intervals $out_dir/result.holdout_intervals.bed \
+    --sizes $ref_dir/hg19.auto.sizes \
```
avantikalal: --sizes does not need to be provided for this command.

ntadimeti:
rerun tests

tests/end-to-end/run.sh
```diff
@@ -0,0 +1,262 @@
...
+env | grep "PATH"
+
+# End-to-end testing script.
+
+verify_diff () {
+    if [ "$3" == "h5" ]
+    then
+        h5diff $1 $2
+    else
+	diff $1 $2
```
tijyojwad: fix indentation to match g5diff line

tests/end-to-end/verify_diff.py
```diff
@@ -0,0 +1,141 @@
...
+        args : parsed argument object.
+
+    """
+    parser = argparse.ArgumentParser(
+        "Utils scripts for verifying identicality of different files.")
+    parser.add_argument('--result_path', type=str,
+                        help='Path to generated result file.')
+    parser.add_argument('--expected_path', type=str,
+                        help='Path to expected result file.')
+    parser.add_argument('--format', type=str,
```
tijyojwad: I don't see a verify function for `bw`. is that not needed anymore?

tests/end-to-end/run.sh
```diff
@@ -0,0 +1,262 @@
...
+# license agreement from NVIDIA CORPORATION is strictly prohibited.
+#
+
+
+set -e
+env | grep "PATH"
+
+# End-to-end testing script.
+
+verify_diff () {
```
tijyojwad: let's move this to the `verify_diff.py` script as well, so we have all forms of verification encapsulated in a single script.

tests/end-to-end/run.sh
```diff
@@ -0,0 +1,262 @@
...
+# Switch to root directory before running script.
+cd $test_root_dir
+
+if [ -d "$out_dir" ]; then
+    rm -rf $out_dir
+fi
+mkdir -p $out_dir
+
+
+echo ""
```
tijyojwad: let's break each of these steps into separate test.sh scripts with descriptive names. so the individual scripts would look like this - 

```
preamble (a shared setup function to set basic directories)
single test, e.g. Split the given regions of the genome into train, val, and holdout/test intervals...
verify diff and output success or failure
```

tests/end-to-end/verify_diff.py
```diff
@@ -0,0 +1,141 @@
...
+        result_path: Path to generated result file.
+        expected_path: Path to expected result file.
+
+    """
+    rank = 0
+    result_model = torch.load(result_path, map_location="cuda:" + str(rank))
+    expected_model = torch.load(expected_path,
+                                map_location="cuda:" + str(rank))
+    for key, _ in result_model['state_dict'].items():
+        val = torch.sum((result_model['state_dict'][key] -
```
tijyojwad: `val` gets overwritten in each loop of the for. did you mean to use `val +=` here?

tests/end-to-end/verify_diff.py
```diff
@@ -0,0 +1,141 @@
...
+#
+"""Test different output files against expected files."""
+
+import argparse
+
+import sys
+
+import torch
+
+EPS = 0.0001
```
tijyojwad: since this is global best to add a prefix to it. something like "VERIFY_DIFF_EPS"

tests/end-to-end/verify_diff.py
```diff
@@ -0,0 +1,141 @@
...
+            if diff > EPS:
+                err_msg = "Metrics for " + key + " differ more than " + str(
+                    EPS) + " between result and expected metrics!"
+                raise ValueError(err_msg)
+
+
+args = parse_args()
+if args.format == "torch":
+    verify_torch(args.result_path, args.expected_path)
+if args.format == "text":
```
tijyojwad: can a user pass both text and torch, or is it one or the other? if it's the latter it would be better to use `if... elif... else` instead of `if... if...`

ntadimeti: done.

tests/end-to-end/verify_diff.py
ntadimeti: done.

tests/end-to-end/run.sh
ntadimeti: removed fn.

tests/end-to-end/verify_diff.py
ntadimeti: removed, good catch.

tests/end-to-end/run.sh
ntadimeti: done.

tests/end-to-end/run.sh
ntadimeti: done.

tests/end-to-end/verify_diff.py
ntadimeti: done.

tests/end-to-end/save_to_h5.sh
```diff
@@ -0,0 +1,67 @@
...
+    --batch_size 64 \
+    --prefix $out_dir/no_label \
+    --nolabel
+
+echo ""
+echo "Verifying output against expected results"
+python $test_dir/verify_diff.py --result_path $out_dir/train_data.h5 \
+	              --expected_path $expected_results_dir/train_data.h5 \
+		      --format "h5"
+if [ $? != 0 ]
```
tijyojwad: let's move this block into a function since it's used everywhere. and the function can take in as an argument a custom informative message to print on exit, e.g. "Failed to match train h5" or "Failed to match eval h5" instead of a more generic "Abort test"

tests/expected_results/end-to-end/reg_log_7a
```diff
@@ -0,0 +1,11 @@
+INFO:2020-02-05 17:29:01,552:AtacWorks-calculate-metrics] Loading intervals
```
tijyojwad: can we rename this file to be more descriptive of the content?

tests/expected_results/end-to-end/reg_log
```diff
@@ -0,0 +1,11 @@
+INFO:2020-01-30 14:36:42,958:AtacWorks-calculate-metrics] Loading intervals
```
tijyojwad: same, rename the file to be more descriptive of the content

tests/end-to-end/verify_diff.py
```diff
@@ -0,0 +1,181 @@
+#!/usr/bin/env python
+
```
tijyojwad: let's move this script to `tests/utils`

tests/end-to-end/infer_and_calculate_metrics.sh
```diff
@@ -0,0 +1,89 @@
...
+if [ $? != 0 ]
+then
+    exit
+fi
+echo "Success!"
+
+echo ""
+echo "Calculate metrics for track coverage after inference..."
+echo ""
+python $root_dir/calculate_baseline_metrics.py \
```
tijyojwad: why not split the metrics calculation into a separate end to end test?

ci/common/test-pyclaragenomics.sh
```diff
@@ -19,6 +19,7 @@ cd $PYCLARAGENOMICS_DIR
 
 #Install external dependencies.
 python -m pip install --ignore-installed -r requirements-base.txt && python -m pip install -r requirements-macs2.txt
+apt install hdf5-tools
```
tijyojwad: this works in CI without using `sudo`?

tests/end-to-end/save_to_h5.sh
tijyojwad: we can use the abstracted function across all tests

tests/end-to-end/train.sh
```diff
@@ -0,0 +1,23 @@
...
+# NVIDIA CORPORATION and its licensors retain all intellectual property
+# and proprietary rights in and to this software, related documentation
+# and any modifications thereto.  Any use, reproduction, disclosure or
+# distribution of this software and related documentation without an express
+# license agreement from NVIDIA CORPORATION is strictly prohibited.
+#
+echo ""
+echo "Train "
+echo ""
+python $root_dir/main.py --train \
```
tijyojwad: we should still check for non zero exit status

tests/expected_results/end-to-end/cla_log
```diff
@@ -0,0 +1,10 @@
+INFO:2020-01-30 16:54:17,661:AtacWorks-calculate-metrics] Loading intervals
```
tijyojwad: rename file with more descriptive name

tests/expected_results/end-to-end/cla_log_7b
```diff
@@ -0,0 +1,10 @@
+INFO:2020-02-05 17:40:04,642:AtacWorks-calculate-metrics] Loading intervals
```
tijyojwad: same, rename file

tests/end-to-end/save_to_h5.sh
ntadimeti: custom messages are printed in python script. So, we don't need specific error messages here.

ci/common/test-pyclaragenomics.sh
ntadimeti: this was not supposed to be here, removed it. Added to the dockerfile.

tests/end-to-end/infer_and_calculate_metrics.sh
ntadimeti: No reason, just inference and metrics seemed to be 1 step to me.

tests/end-to-end/save_to_h5.sh
ntadimeti: Done. verify_diff already prints custom messages. Added a utils.sh script that has a function check_status that can print a custom message too.

tests/expected_results/end-to-end/reg_log_7a
ntadimeti: done.

tests/expected_results/end-to-end/reg_log
ntadimeti: this isn't needed anymore. removed it.

tests/end-to-end/verify_diff.py
ntadimeti: done.

tests/end-to-end/infer_and_calculate_metrics.sh
ntadimeti: done.

tests/end-to-end/train.sh
ntadimeti: agree. removed it by accident. good catch!

tests/expected_results/end-to-end/cla_log
ntadimeti: not needed, removed file.

tests/expected_results/end-to-end/cla_log_7b
ntadimeti: done.

tests/utils/verify_diff.py
```diff
@@ -0,0 +1,182 @@
...
+    parser = argparse.ArgumentParser(
+        "Utils scripts for verifying identicality of different files.")
+    parser.add_argument('--result_path', type=str,
+                        help='Path to generated result file.')
+    parser.add_argument('--expected_path', type=str,
+                        help='Path to expected result file.')
+    parser.add_argument('--format', type=str,
+                        choices=["torch", "text", "h5", "general_diff"],
+                        help='File formats being tested.')
+    parser.add_argument('--metrics', type=str,
```
tijyojwad: what's this option used for?

tests/end-to-end/save_to_h5.sh
tijyojwad: Ah I meant something a little different by the custom messages. verify_diff prints the diff failure between the file names, but sometimes file names get too long and hard to parse. So I was thinking of something like "save to h5 test failed", so in the logs it's easier to follow. we can update it in another PR though. can you file an issue and then resolve this conversation?

ntadimeti: done issue #116

tests/utils/verify_diff.py
ntadimeti: not used anymore, removed.

ohadmo:
rerun tests

PR #112, opened 2020-02-04
fix peak2bw imports and saving

avantikalal:
- import functions from `bedio.py` to read input files in `peak2bw.py`
- simplify `read_sizes` function in `bedio.py`
- sort bedGraph file written by `peak2bw.py` - needed for bigWig conversion.
- check bedgraph output of `peak2bw.py` against chromosome sizes file before writing


claragenomics/io/bedio.py
```diff
@@ -14,11 +14,12 @@
 import pandas as pd
 
 
-def read_intervals(bed_file):
-    """Function to read genomic intervals from a BED file into a DataFrame.
+def read_intervals(bed_file, skip=0):
```
ntadimeti: Why do we need skip option ?

claragenomics/io/bedio.py
```diff
@@ -27,24 +28,27 @@ def read_intervals(bed_file):
...
+                     dtype={'chrom': str, 'length': int})
     if as_intervals:
-        df[2] = [0] * len(df)
-        df.rename(columns={0: 0, 2: 1, 1: 2}, inplace=True)
-        df.columns = ['chrom', 'start', 'end']
-    else:
-        df.columns = ['chrom', 'length']
+        df['start'] = [0] * len(df)
+        df = df[['chrom', 'start', 'length']]
+        df.rename(columns={"length": "end"}, inplace=True)
```
ntadimeti: why is length being renamed to end ? May be leave a comment for this operation, so we are not confused later on.

peak2bw.py
```diff
@@ -70,19 +70,18 @@ def main():
...
     # Read input files
     _logger.info('Reading input files')
-    peaks = pd.read_csv(args.input, sep='\t', header=None,
-                        usecols=[0, 1, 2], skiprows=args.skip)
-    sizes = pd.read_csv(args.sizes, sep='\t', header=None)
+    peaks = read_intervals(args.input, skip=args.skip)
+    sizes = read_sizes(args.sizes)
 
     # Subset to peaks in sizes file
     # TODO: Move this test into the df_to_bedGraph function
```
ntadimeti: is this TODO still relevant ? If so, can we make this change in this PR and remove the TODO ?

peak2bw.py
```diff
@@ -70,19 +70,18 @@ def main():
...
-    _logger.info('Retaining ' + str(len(peaks_filtered)) + ' of ' +
-                 str(len(peaks)) + ' peaks in given chromosomes.')
+    peaks_filtered = peaks[peaks['chrom'].isin(sizes['chrom'])].copy()
+    _logger.info('Retaining ' + str(len(peaks_filtered)) + ' of ' + str(len(
+        peaks)) + ' peaks in given chromosomes.')
 
     # Add score of 1 for all peaks
     _logger.info('Adding score')
-    peaks_filtered[3] = 1
+    peaks_filtered['score'] = 1
```
ntadimeti: do we not need a column of 1st equal to the `len(peaks_filtered)` ?

avantikalal: Sorry, I didn't understand this comment. What do you mean by 'column of 1st'?

claragenomics/io/bedio.py
avantikalal: In case we're reading an interval file with a header. Currently used by `peak2bw.py`.

claragenomics/io/bedio.py
avantikalal: Added comments.

avantikalal: It's done so that the output has the same format as the output of `read_intervals`. So that the same downstream function can take input either from `read_intervals` or `read_sizes`. Currently we use this in `calculate_baseline_metrics.py`.

peak2bw.py
avantikalal: Made the change.

peak2bw.py
avantikalal: peaks already has 3 columns: 'chrom', 'start' and 'end', and we're adding a 4th column which is 'score'.

claragenomics/io/bedio.py
ntadimeti: can we replace "length" with "end" in line 48. That way we don't have to rename it.

peak2bw.py
ntadimeti: Sorry, I meant column of ones. `peaks_filtered['score'] = 1` What would be the `len(peaks_filtered['score'] = 1)` does pandas assign it automatically ?

avantikalal: Pandas assigns a 1 to the score column in every row of the dataframe.

claragenomics/io/bedio.py
avantikalal: Not straightforward. In both main.py and df_to_bedGraph, we have to merge the two dataframes (intervals and sizes) together. The merge won't work if they have the same column name ('end') so I would have to rename one of them anyway. Better to have different names from the beginning. Also in my opinion it's confusing to call a column 'end' in a dataframe where there is no column named 'start'.

PR #111, opened 2020-02-03
Tutorials

avantikalal:
Tutorial 1: Training
Tutorial 2: Inference

PR #121, opened 2020-02-13
Add setup.py

ntadimeti:


setup.py
```diff
@@ -0,0 +1,123 @@
...
+      install_requires=[get_installation_requirments(get_verified_absolute_path(os.path.join(current_dir, 'requirements-base.txt'))),
+                        get_installation_requirments(get_verified_absolute_path(os.path.join(current_dir, 'requirements-macs2.txt')))
+                       ],
+      packages=find_namespace_packages(where=current_dir, include=['claragenomics.*']),
+      python_requires='>=3.5',
+      license='Apache License 2.0',
+      long_description='Python libraries and utilities for manipulating genomics data',
+      classifiers=pyaw_classifiers,
+      platforms=['any'],
+      scripts=[os.path.join('scripts', 'bw2h5.py'),
```
tijyojwad: let's remove these scripts for now so they don't get installed to the conda `bin` folder

setup.py
```diff
@@ -0,0 +1,123 @@
...
+      description='NVIDIA genomics python libraries and utiliites',
+      author='NVIDIA Corporation',
+      url="https://github.com/clara-genomics/AtacWorks",
+      include_package_data=True,
+      install_requires=[get_installation_requirments(get_verified_absolute_path(os.path.join(current_dir, 'requirements-base.txt'))),
+                        get_installation_requirments(get_verified_absolute_path(os.path.join(current_dir, 'requirements-macs2.txt')))
+                       ],
+      packages=find_namespace_packages(where=current_dir, include=['claragenomics.*']),
+      python_requires='>=3.5',
+      license='Apache License 2.0',
```
tijyojwad: let's remove this for now since we have a custom license for AtacWorks

setup.py
```diff
@@ -0,0 +1,123 @@
...
+
+# Get current dir (pyclaragenomics folder is copied into a temp directory created by pip)
+current_dir = os.path.dirname(os.path.realpath(__file__))
+
+# Classifiers for PyPI
+pyaw_classifiers = [
+        "Development Status :: 5 - Production/Stable",
+        "Intended Audience :: Science/Research",
+        "Topic :: Scientific/Engineering :: Bio-Informatics",
+        "License :: OSI Approved :: Apache Software License",
```
tijyojwad: Let's remove this license as well

setup.py
```diff
@@ -0,0 +1,123 @@
...
+    Returns:
+        list of requirement strings
+    """
+    with open(file_path, 'r') as file:
+        requirements_file_content = \
+            [line.strip() for line in file if line.strip() and not line.lstrip().startswith('#')]
+    return requirements_file_content
+
+
+def copy_all_files_in_directory(src, dest, file_ext="*.so"):
```
tijyojwad: this function is not used

setup.py
```diff
@@ -0,0 +1,123 @@
...
+        for file in files_to_copy:
+            shutil.copy(file, dest)
+            print("{} was copied into {}".format(file, dest))
+    except (shutil.Error, PermissionError) as err:
+        print('Could not copy {}. Error: {}'.format(file, err))
+        raise err
+
+
+# Must be set before calling pip
+for envvar in ['AW_INSTALL_DIR', 'AW_VERSION']:
```
tijyojwad: this section is not needed since it's specific to the way CGA setup.py is called

setup.py
```diff
@@ -0,0 +1,123 @@
...
+        "Operating System :: POSIX :: Linux",
+        "Programming Language :: Python :: 3.5",
+        "Programming Language :: Python :: 3.6",
+        "Programming Language :: Python :: 3.7",
+        "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9"
+]
+
+setup(name='atacworks',
+      version=aw_version,
```
tijyojwad: for now we can hard code the version numbers here. this is a holdover from CGA setup

tijyojwad: can you add the version number directly to the variable in `setup` call instead of using the other variable? we can add it later if we derive from a variable

ntadimeti: done.

ntadimeti:
@avantikalal just FYI about this. It moves all the scripts under scripts directory. I will update the #111 tutorials PR once this is merged. 

You have to run `pip install .` from Atacworks root dir and then you can will be able to use the scripts same as before.

README.md
```diff
@@ -69,6 +69,11 @@ git clone --recursive https://github.com/clara-genomics/AtacWorks.git
     pip install -r requirements-base.txt && pip install -r requirements-macs2.txt
     ```
 
+* Install atacworks
```
ntadimeti: @tijyojwad does this look good ?

tijyojwad: looks good!

PR #128, opened 2020-02-24
Add subparsers to main.py

ntadimeti:
Organize command line arguments into sub-commands train, infer and eval. 

There are some common arguments for all modes, those should be specified first, followed by the mode of operation, followed by arguments specific to the mode of operation. 
```python scripts/main.py -h gives the following help : 
usage: main.py [-h] {train,infer,eval} ...

positional arguments:
  {train,infer,eval}

optional arguments:
  -h, --help          show this help message and exit
```

python scripts/main.py $mode -h prints out the command line arguments specific to the $mode. 

```python scripts/main.py train -h 
usage: main.py train [-h] --label LABEL --out_home OUT_HOME --task
                     {regression,classification,both} --print_freq PRINT_FREQ
                     --bs BS --num_workers NUM_WORKERS --pad PAD --transform
                     {log,None} [--layers LAYERS] --weights_path WEIGHTS_PATH
                     --gpu GPU [--distributed] --dist-url DIST_URL
                     --dist-backend DIST_BACKEND [--debug] --files_train
                     FILES_TRAIN --checkpoint_fname CHECKPOINT_FNAME
                     --save_freq SAVE_FREQ --clip_grad CLIP_GRAD --lr LR
                     --epochs EPOCHS --mse_weight MSE_WEIGHT --pearson_weight
                     PEARSON_WEIGHT --poisson_weight POISSON_WEIGHT
                     --val_files VAL_FILES --eval_freq EVAL_FREQ --threshold
                     THRESHOLD --best_metric_choice
                     {BCE,MSE,Recall,Specificity,CorrCoef,AUROC}

```


tijyojwad:
@ntadimeti is the order of options you specified above (having some options before the mode and some after) a requirement?

ntadimeti:
> @ntadimeti is the order of options you specified above (having some options before the mode and some after) a requirement?

With current implementation, yes.
The other option would be to replicate all of these common options for every mode, which would be cumbersome. 
There's another option where we can inherit the common parser as a parent for each of these sub parsers. I tried that initially, there's some issues with conflicting help messages. We will have to turn off the help for default parser, which means when people do `python main.py -h` they get no help. They will have to know apriori what modes exist.

Happy to learn more

tijyojwad:
maybe an option is to have a helper function that adds the common options to a parser, so cmd_args.py can have something that looks like - 

```
def add_common_options(parser):
    parser.add_argument(option1)
    parser.add_argument(option2)
```

and each of the sub parsers for train, infer, eval get assed through this function. This way we limit repetitive code, but also duplicate the options across all parsers.

A nicer thing would be to add these to an abstract parser class that train/infer/eval inherit from, but it looks like inheritance is messing with the default parser? The abstract parser should also be inheritable from the default parser, so my gut feeling is it shouldn't break anything. But the common function should work

ntadimeti:
@avantikalal --FYI. 

Everything works just as before. Instead of `main.py --train` you will now just provide `main.py train`

scripts/worker.py
```diff
@@ -144,10 +144,14 @@ def get_model(args, gpu, rank):
                                 world_size=args.world_size, rank=rank)
 
     # Why is model & optimizer built in spawned function?
+    resume = False
+    if args.weights_path is not None:
```
tijyojwad: `resume = (args.weights_path is not None)` should do it

this resume functionality is also going to be fixed in another PR right?

scripts/cmd_args.py
```diff
@@ -64,191 +64,210 @@ def type_or_none_fn(type):
...
+    add_inference_options(parser_infer)
+    # =========================================================================
+    # Evaluation args
+    parser_eval = subparsers.add_parser(
+        'eval',
+        config_file_parser_class=configargparse.YAMLConfigFileParser,
+        default_config_files=[infer_config_path])
+
+    add_common_options(parser_eval)
+    add_inference_options(parser_eval)
```
tijyojwad: can we add an `add_train_options` and `add_eval_options` function as well? and then those functions can call `add_common_options` from inside them?

scripts/worker.py
ntadimeti: yes, created an issue for this  #130

scripts/cmd_args.py
ntadimeti: done

scripts/worker.py
tijyojwad: Perfect, thanks!

PR #141, opened 2020-03-23
Prefetch layers in inference to avoid OOM issues

ntadimeti:
Need to add test for layers feature. Waiting for Avantika to test on her experiments and make sure everything looks good.

### DIFF
--- a/atacworks/dl4atac/dataset.py
+++ b/atacworks/dl4atac/dataset.py
@@ -33,7 +33,7 @@ class DatasetBase(Dataset):
         self.files = files
         self.layers = layers
         self._h5_gen = None
-        assert len(files) > 0,\
+        assert len(files) > 0, \
             "Need to supply at least one file for dataset loading"
         self.running_counts = [0]
         for file in self.files:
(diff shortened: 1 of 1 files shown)


claragenomics/dl4atac/dataset.py
```diff
@@ -170,24 +170,31 @@ def _get_generator(self):
...
             # Find correct dataset, given idx
             file_id, local_idx = self._get_file_id(idx)
+            filename = self.files[file_id]
             assert file_id < len(self.files)
             if (local_idx < self.fh_indices[file_id][0]) or (
                     local_idx >= self.fh_indices[file_id][1]):
                 # Data is not pre loaded, so load new data
+                # Read additional layers
+                hf = h5py.File(filename, 'r')
+                prefetch_hdrecs = hdrecs[file_id][
```
tijyojwad, line 186: since the other layers are still being loaded from a file here, I don't see much of a point in loading the `input` layer in line 172 anymore. It's probably best to refactor it as the following - 
1. near line 170, just open the h5py file and store the file pointer (so we don't have to keep closing and opening it), and also update the "layers" list to have input + custom layer
2. near line 184, porefetch all layers and stack them up

ntadimeti: In line 170, if I do 
```
hdrecs.append(hf)```
instead of 
```
 hdrecs.append(hf["input"])```

program still crashes with OOM. I'm not sure how to save file pointer, as reading the h5 file stores the entire file object. Am I mis-understanding something ?

tijyojwad: Strange... In that case I'd suggest we just move the whole File reading down into line 184 as well, and take out the loop that's there near line 170. So 184 would look like this
```
hf = h5py.File(self.files[file_id], 'r')
prefetch_hdrec = hf[input]local_index + prefetch_size]
for layer in layers:
    prefetch_hdrec[layers] = np.stackprefetch_hdrec , (hf[layer]local_index + prefetch_size])
```

PR #143, opened 2020-04-01
Default to gpu 0 when requested index is more than number of devices

ntadimeti:
Addresses this issue : https://github.com/clara-genomics/AtacWorks/issues/139

### DIFF
--- a/scripts/main.py
+++ b/scripts/main.py
@@ -355,10 +355,16 @@ def main():
             mp.spawn(train_worker, nprocs=ngpus_per_node,
                      args=(ngpus_per_node, args), join=True)
         else:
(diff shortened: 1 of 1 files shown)


tijyojwad:
It's probably not a good idea to default to GPU 0 if the user specified GPU is unavailable. Mainly because the user likely specified a GPU for a reason and GPU 0 may be running other tasks which this run could interfere with.

I would prefer to just throw an error and stop (like the current setup).

ntadimeti:
@avantikalal  see Joyjit's comment above. Does that sound okay to you? will close the PR if so.

avantikalal:
Yes, makes sense.

Issue #145, opened 2020-05-10
Exception: Invalid format for file paths provided


wookyung:
Hello, 

Thanks for making this tool! 
While I was testing Tutorial 1 with your docker version, I got the following error at step 7. I checked all paths are properly set. Any idea or hint to fix this error? Thanks!

```
$atacworks/scripts/main.py train --config configs/train_config.yaml --config_mparams configs/model_structure.yaml --train_files Mono.50.2400.train.h5 --val_files Mono.50.2400.val.h5

Traceback (most recent call last):
  File "/AtacWorks/scripts/main.py", line 443, in <module>
    main()
  File "/AtacWorks/scripts/main.py", line 331, in main
    args.files_train = gather_files_from_cmdline(args.files_train)
  File "/usr/local/lib/python3.6/dist-packages/claragenomics/dl4atac/utils.py", line 234, in gather_files_from_cmdline
    raise Exception("Invalid format for file paths provided.")
Exception: Invalid format for file paths provided.
```

ntadimeti:
Hi @wookyung 

Looks like it's a typo in our tutorial. Good catch. Can you change `--train_files` to `files_train`. It should work fine. I will create a PR to fix this immediately. Thank you.

wookyung:
Thanks for trying to fix it but I got the same errors with --files_train/--files_val. Was the fixing applied to a docker version? Thanks!

```
$atacworks/scripts/main.py train --config train_config.yaml --config_mparams model_structure.yaml --files_train Mono.50.2400.train.h5 --files_val Mono.50.2400.val.h5
Traceback (most recent call last):
  File "/AtacWorks/scripts/main.py", line 443, in <module>
    main()
  File "/AtacWorks/scripts/main.py", line 332, in main
    args.val_files = gather_files_from_cmdline(args.val_files)
  File "/usr/local/lib/python3.6/dist-packages/claragenomics/dl4atac/utils.py", line 234, in gather_files_from_cmdline
    raise Exception("Invalid format for file paths provided.")
Exception: Invalid format for file paths provided.
```

wookyung:
Hi @ntadimeti ,

I tried it again with --files_train and --val_files and got the more complicated errors as below:

```
$atacworks/scripts/main.py train --config train_config.yaml --config_mparams model_structure.yaml --files_train Mono.50.2400.train.h5 --val_files Mono.50.2400.val.h5
[...]

PR #146, opened 2020-05-11
Fix typo in training

ntadimeti:
`train_files` has been changed to `files_train` to avoid argument parsing errors. train_files was being parsed as args.train, which is an already existing argument. This change is not reflected in the tutorial1.


ntadimeti:
Addresses #145

tijyojwad:
@ntadimeti let's also merge this as a hotfix to master branch and update the patch version number?

ntadimeti:
sure, we can do that.

PR #147, opened 2020-05-11
Fix typo in training

ntadimeti:


tijyojwad:
@ntadimeti - can you also update the patch number in the version in `setup.py`?

ntadimeti:
@tijyojwad done.

Issue #98, opened 2020-01-14
main.py : Function to set random seed. 


ntadimeti:
The function to add random seed is currently not being used anywhere and we are not sure it is functional. Removing it from main.py for now. Validate and add this feature back.

```def set_random_seed(seed):
    """Set random seed for reproducability."""

    if seed is not None and seed > 0:
        random.seed(seed)
        np.random.seed(seed)
        torch.manual_seed(seed)
        mp.model_parallel_cuda_manual_seed(seed)
```

ntadimeti:
Currently AtacWorks uses certain pytorch functions that do not guarantee determinism. All the other parts of the code that can be deterministic, utitlize the seed. Issue #150 addresses this.

PR #150, opened 2020-05-13
Set seed

ntadimeti:
Addresses #98 

Added option to set a seed for all random operations. This however, doesn't make the model deterministic.

### DIFF
--- a/scripts/worker.py
+++ b/scripts/worker.py
@@ -35,7 +35,6 @@ import torch
 import torch.distributed as dist
 
 from torch.optim import Adam
-
 warnings.filterwarnings("ignore")
 
 # Set up logging
@@ -165,6 +164,10 @@ def train_worker(gpu, ngpu_per_node, args, timers=None):
         timers : .
 
     """
+    # fix random seed so models have the same starting weights
+    if args.seed is not None and args.seed > 0:
+        torch.manual_seed(args.seed)
+        torch.cuda.manual_seed(args.seed)
     rank = gpu if args.distributed else 0
 
     model, model_params = get_model(args, gpu, rank)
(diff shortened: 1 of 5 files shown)


scripts/worker.py
```diff
@@ -166,7 +166,8 @@ def train_worker(gpu, ngpu_per_node, args, timers=None):
 
     """
     # fix random seed so models have the same starting weights
-    torch.manual_seed(42)
+    if args.seed is not None and args.seed > 0:
+        torch.manual_seed(args.seed)
```
tijyojwad, line 170: Since torch.manual_seed is set in main, is this needed again?

ntadimeti: I'm not sure if we need to set it everytime torch is imported. :-/ I didn't find much documentation supporting either theories. I left it in there for safety. Happy to learn more.

tijyojwad: I'm not sure either. Would be good to find out. Can probably also try to fetch the manual seed to see if it stays fixed in different parts of the code. The fewer place this has to be called the safer it is. My hunch is it might not be for each import, but might be once for each process

ntadimeti: I'm not sure there's an API to do that, didn't find any in the pytorch documentation

ntadimeti: re-setting the seed in the workers is not affecting the determinism of the models anyway. The models are still non-deterministic (binary files differ). So, I can remove them for now and when we investigate about determinism, we can think more about it. sound good ?

ntadimeti: @tijyojwad removed the seed setting fro mworkers. Added cuda.manual_seed() also in scripts/main.py.

tijyojwad: Thanks @ntadimeti !

scripts/worker.py
```diff
@@ -165,9 +164,9 @@ def train_worker(gpu, ngpu_per_node, args, timers=None):
         timers : .
 
     """
-    # fix random seed so models have the same starting weights
-    torch.manual_seed(42)
-
+    if args.seed is not None and args.seed > 0:
```
tijyojwad, line 167: @ntadimeti approving, but can you add this comment back to `train_worker`? 
```
    # fix random seed so models have the same starting weights
```

PR #152, opened 2020-05-13
Update command line options and tutorials

ntadimeti:
```
val_files -> files_val
train_files -> files_train
sizes_files -> files_chromsizes
intervals_file -> file_intervals

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,7 @@ The model learns a mapping from (1) to both (2) and (3); in other words, from th
 
 #### Tutorial
 
+NOTE: Tutorials are not compatible across different branches. Currently tutorials are only compatible with the master branch.
 See [Tutorial 1](tutorials/tutorial1.md) for a workflow detailing the steps of data processing, encoding and model training and how to modify the parameters used in these steps.
 
 ### 2. Denoising and peak calling using a trained AtacWorks model
--- a/setup.py
+++ b/setup.py
@@ -66,7 +66,7 @@ pyaw_classifiers = [
 ]
 
 setup(name='atacworks',
-      version='0.2.2',
+      version='0.3.0',
       description='NVIDIA genomics python libraries and utiliites',
       author='NVIDIA Corporation',
       url="https://github.com/clara-genomics/AtacWorks",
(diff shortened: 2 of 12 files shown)

### CODE REVIEW

File: setup.py [hunk 1]
```diff
@@ -66,7 +66,7 @@ def get_installation_requirments(file_path):
 ]
 
 setup(name='atacworks',
-      version='0.2.0',
+      version='0.2.2',
```
tijyojwad, line 69: Is this PR supposed to be for master or dev 0.3.0? If for dev 0.3.0, then we should just update the version to be 0.3.0

tijyojwad:
@ntadimeti another thing we should clarify in README that API in master branch could be different from API in development branch. So the configs from AWS may not work with both. Or if we make a breaking change then we should specify that tutorial only work for master branch

setup.py
ntadimeti: done.

ntadimeti:
@tijyojwad  added a note in the readme about compatibility of tutorials with master branch

README.md
```diff
@@ -87,6 +87,7 @@ The model learns a mapping from (1) to both (2) and (3); in other words, from th
 
 #### Tutorial
 
+NOTE: Tutorials are not compatible across different branches. Currently tutorials are only compatible with the master branch.
```
tijyojwad, line 90: is the tutorial only compatible with `master` branch or with the branch it belong to? e.g. the changes in this PR update the tutorials to use the new args, but those will only be available in `dev-v0.3.0` branch right?

ntadimeti: It's only compatible with `master` branch, because we have a copy of configs uploaded on S3 which are compatible with master branch. Even though the tutorials.md files here are updated, they aren't yet compatible with uploaded data. 

We should begin some sort of versioning of the tutorials data and then we can update the wget links to fetch proper versions. I could create a folder dev-0.3.0 on S3, but I wanted to figure out a better startegy and then have a PR that updates it.

tijyojwad: I like the tutorial versioning idea. Let's figure that one out asap. The issue with the changes in this PR is that the tutorial says it only works with master branch, but then it uses options that are newly added to v0.3.0. I think this can be confusing to the user. With versioning that will absolutely get better

ntadimeti:
Closing this PR since a lot is changing wrt to APIS and tutorials. This is obsolete.

PR #154, opened 2020-05-14
updated output folder name in tutorial 1

avantikalal:


ntadimeti:
can you also do the following 
1. change all `train_files` to `files_train`. I merged a PR that only updates one command, I didn't realize there were many train commands.
2. Tutorial1 & 2 : Add step to download all data files into a new folder `data` or something. This will prevent users from using default configs and running into issues.

tutorials/tutorial1.md
```diff
@@ -41,7 +46,7 @@ wget https://atacworks-paper.s3.us-east-2.amazonaws.com/dsc_atac_blood_cell_deno
 ### Config files
 We also need to download config files for this experiment. The config files describe the structure of the deep learning model and the parameters to train it. We will place these in the `configs` folder. 
 ```
-mkdir configs
+mkdir config_files
```
ntadimeti, line 49: This change is not needed when inside a new dir.

ntadimeti:
Testing the tutorial1 and 2 instructions. Will update here

tutorials/tutorial1.md
```diff
@@ -153,10 +158,10 @@ To train the model, we supply the training and validation datasets as well as th
 python $atacworks/scripts/main.py train \
         --config configs/train_config.yaml \
```
ntadimeti, line 159: For the config files, let's use absolute path.

tutorials/tutorial1.md
```diff
@@ -153,10 +158,10 @@ To train the model, we supply the training and validation datasets as well as th
...
         --config_mparams configs/model_structure.yaml \
```
ntadimeti, line 160: let's use absolute path: `$atacworks/tutorial1/configs/model_structure.yaml`

ntadimeti:
There seems to be a bug in the parser itself, it's not picking up any custom configs. I will create a PR to look into that and add appropriate tests.

tutorials/tutorial1.md
avantikalal: This would mean that the tutorial has to be run in the AtacWorks folder. I'm not sure this is a good idea, users should be able to download the tutorial files anywhere.

ntadimeti: Okay makes sense. I'll then figure this out in the next PR. Let's leave this as is.

PR #158, opened 2020-05-22
Pick up custom config files

ntadimeti:
Code now picks up custom config files. Copy paste error. Verifyong tutorial 1 to make sure it works as expected.


ntadimeti:
rerun tests

configs/infer_config.yaml
```diff
@@ -40,3 +40,7 @@ batches_per_worker: 16
 result_fname: inferred
 gen_bigwig: True
 deletebg: False
+
+#Eval options
+best_metric_choice: "AUROC"
```
avantikalal, line 45: Why do we need best_metric during inference?

ntadimeti: We need for evaluation. Not for inference. `python main.py eval` needs it. This is in addition to all the options required by infer. It didn't make sense to have yet another config file, so I added these options in the infer_config.

ntadimeti:
rerun tests

Issue #153, opened 2020-05-14
main.py infer: error: argument --gpu: invalid int value: 'None'


wookyung:
Hello, 

I am sorry but I don't know where/whom I can ask, so just leave some questions again here (It would be great for me to know the contact info like email address). I got some issues while I was trying both tutorial 1 and 2 as below:

1) Here's the log when I ran the following command at step7 of tutorial 1. It just finished without any error or warning. I don't still get final files that you expect after it's done. Is is still because of inappropriate memory or gpu setting (please see the following setting)? 

```
export LSF_DOCKER_NETWORK=host
export LSF_DOCKER_IPC=host
export LSF_DOCKER_SHM_SIZE=3g
bsub -G compute-yooa -Is -q general-interactive -gpu "num=4:gmodel=TeslaV100_SXM2_32GB" -R 'rusage[mem=64GB]' -M 64GB -a 'docker(claraomics/atacworks)' /bin/bash
```

```
$atacworks/scripts/main.py train --config train_config.yaml --config_mparams model_structure.yaml --files_train Mono.50.2400.train.h5 --val_files Mono.50.2400.val.h5

INFO:2020-05-13 16:37:02,912:AtacWorks-main] Running on GPU: 0
Building model: resnet ...
Finished building.
Saving config file to ./trained_models_2020.05.13_16.37/configs/model_structure.yaml...
Num_batches 500; rank 0, gpu 0
Epoch [ 0/25] -------------------- [  0/500] mse:  20.142 | pearsonloss:   0.986 | total_loss:   1.603 | bce:   0.607
Epoch [ 0/25] ##------------------ [ 50/500] mse:2030.479 | pearsonloss:   0.055 | total_loss:   1.578 | bce:   0.507
Epoch [ 0/25] ####---------------- [100/500] mse:  20.136 | pearsonloss:   0.986 | total_loss:   1.079 | bce:   0.083
Epoch [ 0/25] ######-------------- [150/500] mse:1936.060 | pearsonloss:   0.015 | total_loss:   1.054 | bce:   0.071
Epoch [ 0/25] ########------------ [200/500] mse:  20.369 | pearsonloss:   0.986 | total_loss:   1.109 | bce:   0.112
Epoch [ 0/25] ##########---------- [250/500] mse:  58.329 | pearsonloss:   0.012 | total_loss:   0.100 | bce:   0.058
[...]

PR #166, opened 2020-06-04
invalid_crossdevice_link

ntadimeti:
This PR addresses one of the issues being tracked in #153 . User is getting the following error when running atacworks inference:
```
Inference -------------------- [ 0/29]

Inference time taken:  120.692s (Load    3.292s,Prediction  117.140s)

INFO:2020-05-15 02:53:50,337:AtacWorks-main] Waiting for writer to finish...

Process Process-2:

Traceback (most recent call last):

  File "/usr/lib/python3.6/shutil.py", line 550, in move

    os.rename(src, real_dst)

OSError: [Errno 18] Invalid cross-device link:
'/tmp/247693.tmpdir/tmptkst0za1/0/00001' ->
'./inference_output_2020.05.15_02.51/NK_inferred.track.bedGraph'
```
Although it's not reproducible on our end, the error seems to be because `os.rename` and `shutile.move` fail when source and destinations seem to be on different file systems. 

An easy workaround for that is to use `shutil.copy` and os.remove`. Inspired from this thread : https://github.com/ewels/MultiQC/issues/806

-- Will wait for user to verify and report back that it works before merging.

### DIFF
--- a/scripts/main.py
+++ b/scripts/main.py
@@ -268,7 +268,7 @@ def writer(infer, intervals_file, exp_dir, result_fname,
             # Only one file should be left after the combiner stage
             assert (len(files) == 1)
             # move final files out of tmp folder
-            shutil.move(files[0], outfiles[channel])
+            shutil.copy(files[0], outfiles[channel])
 
     # remove tmp folder
     shutil.rmtree(temp_dir)


scripts/main.py
```diff
@@ -268,7 +268,7 @@ def writer(infer, intervals_file, exp_dir, result_fname,
             # Only one file should be left after the combiner stage
             assert (len(files) == 1)
             # move final files out of tmp folder
-            shutil.move(files[0], outfiles[channel])
+            shutil.copy(files[0], outfiles[channel])
```
tijyojwad, line 271: @ntadimeti should there be an `os.remove` here as well?

ntadimeti: we have a os.rmtree right after the loop. Didn't find it necessary to have os.remove explicitly ?

tijyojwad: Cool, I saw the rmtree on the tempdir but wasn't sure if it removed the same folder as the move operation. Looks like it does! Resolving.

ntadimeti:
@tijyojwad @avantikalal -- should we merge this into master or dev-0.3.0 ?

PR #174, opened 2020-06-18
CI scripts cleanup

ohadmo:


ntadimeti:
lgtm

ntadimeti:
Let's wait for @tijyojwad

ci/common/prep-init-env.sh
```diff
@@ -19,21 +19,32 @@ logger "Check versions..."
...
 ################################################################################
 
-CUDA_REL=${CUDA:0:3}
-if [ "${CUDA:0:2}" == '10' ]; then
-  # CUDA 10 release
-  CUDA_REL=${CUDA:0:4}
+# FIX Added to deal with Anancoda SSL verification issues during conda builds
+conda config --set ssl_verify False
+
+logger "Activate anaconda environment..."
```
ohadmo, line 34: Since `gdf` anaconda env was not activated, we were testing in gpuCI with python 3.7

requirements-base.txt
```diff
@@ -14,7 +14,7 @@ h5py==2.9.0
 numpy==1.17.0
 pandas==0.25.0
 pyBigWig==0.3.17
-pytest==4.4.1
+pytest>=5.4.3
```
ohadmo, line 17: Received an internal Pytest ImpotError when running this version with the correct anaconda env for Python 3.6.

ci/cpu/build.sh
```diff
@@ -9,36 +9,38 @@
...
+logger "Calling prep-init-env..."
+source ci/common/prep-init-env.sh
 
 ################################################################################
-# Pyclaragenomics tests
+# Install AtacWorks
 ################################################################################
+logger "Insalling AtacWorks..."
+cd "${WORKSPACE}"
+source ci/common/install-package.sh
```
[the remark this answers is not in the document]
ohadmo: @tijyojwad The reason I separated these two steps is for the next PR, so we don't duplicate code when running the tests after creating and installing `.whl` file.
These two `builds.sh` scripts are going to be different -  `.whl file` and docker image will be created only in the CPU job which will run after the GPU job (only installing from source ) finishes successfully.

PR #177, opened 2020-06-29
Change --gpu to --gpu_id to clarify the functionality

ntadimeti:


vellamike:
please don't use underscore in longopts, it can get eaten by terminal borders and look like a space, longopts should be `--like-this`

ntadimeti:
@vellamike thanks for the input! I'll change it accordingly

ntadimeti:
@vellamike --  I saw your comment again. We don't have short option and long options. So, we actually access this argument with `args.gpu_idx` and a hiphen isn't compatible in that case. Am I making sense? Happy to discuss offline

PR #178, opened 2020-07-01
Alal/write binned

avantikalal:
1. Allows binning of Atacworks output (currently by taking the average score over a bin). This saves time by writing at lower resolution and allows AtacWorks output to be used in workflows which expect binned files.

2. Corrected an error in writing `args.interval_size` to output `infer_config.yaml` - padding should be subtracted in calculating the interval size.


ntadimeti:
Have you tested this ? @avantikalal . Can we add a quick end-to-end test for this feature ?

PR #179, opened 2020-07-02
PYPI installation script

ntadimeti:
Setup script for PYPI installation. Currently commented out the installation steps in the script. Once everything looks good, we can create a PR to uncomment it and then do the upload.


Dockerfile
```diff
@@ -27,8 +27,5 @@ RUN rsync -aP rsync://hgdownload.soe.ucsc.edu/genome/admin/exe/linux.x86_64/bedG
 RUN rsync -aP rsync://hgdownload.soe.ucsc.edu/genome/admin/exe/linux.x86_64/bigWigToBedGraph /usr/local/bin
 
 # Download AtacWorks repo
-RUN git clone --recursive -b master https://github.com/clara-genomics/AtacWorks.git
-
-# Install AtacWorks requirements
-RUN pip3 install -r AtacWorks/requirements-base.txt && pip3 install -r AtacWorks/requirements-macs2.txt
-RUN cd AtacWorks && pip3 install --upgrade setuptools && pip3 install .
+RUN pip3 install -i https://test.pypi.org/simple/  --extra-index-url https://pypi.org/simple atacworks==0.2.2
```
tijyojwad, line 30: what's the reason we want to upload to test PyPI repo instead of production PyPI repo?

ntadimeti: Good catch. Ideally this should be production pypi. Since we haven't done that yet, I have some test versions on test.pypi server and was experimenting with that. We can change it to production PYPI once we upload it there. Thoughts? 

Also, I am able to succesfully install and run atacworks on my local machine, but encountering some issues when running this Dockerfile. Will only merge once I can run it. Wanted to put this up for visibility and feedback

ntadimeti:
@tijyojwad having the test.pypi in CI now, will help us test that upload works through CI as well.

Dockerfile
tijyojwad: Yeah sounds good. For final users they should pull from official PyPI repo

tijyojwad:
> 
> 
> @tijyojwad having the test.pypi in CI now, will help us test that upload works through CI as well and then we can change it to the production URL

Will this upload everytime we test? @ohadmo to comment on uploading from tests to see if it could cause any issues on PyPI side wrt version numbering and so on

Dockerfile
ohadmo: In order not to "burn" the version number in PyPI production repo

ci/common/test-atacworks.sh
```diff
@@ -20,3 +20,6 @@ if [ "${TEST_ON_GPU}" == '1' ]; then
 else
     python -m pytest -m cpu tests/ -v
 fi
+
+# Install a wheel package
```
ohadmo, line 24: @ntadimeti `create` instead of `install`

Also, this command should be executed when running a CPU job, not a GPU job. We should execute `test-atacworks.sh` twice in the CPU job- for the build from source code and after installing the `.whl` file we created.

Dockerfile
ntadimeti: Done.

Dockerfile
```diff
@@ -11,9 +11,9 @@
 # Inherit from base NVIDIA CUDA 10.1 image with CUDNN
 FROM nvidia/cuda:10.1-cudnn7-runtime-ubuntu18.04
 
-# Set python version to 3.6
+# Set python version to 3.7
 RUN apt update && apt install -y \
-    python3.6 \
+    python3.7 \
```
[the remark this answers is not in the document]
ntadimeti: I didn't mean to push this, reverted the change.

ci/common/test-atacworks.sh
ntadimeti: Done

Dockerfile
ntadimeti: done.

Dockerfile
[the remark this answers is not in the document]
ntadimeti: this change is not needed. reverted it.

ci/common/test-atacworks.sh
ntadimeti: done.

ntadimeti:
rerun tests

ci/cpu/build.sh
```diff
@@ -44,3 +44,32 @@ source ci/common/install-package.sh
 logger "Running AtacWorks tests..."
 cd "${WORKSPACE}"
 source ci/common/test-atacworks.sh "$WORKSPACE"
+
```
ohadmo, line 47: @ntadimeti we should add here a step to `pip uninstall` the currently installed version of atacworks (develop mode). Since the `.whl` package and the source files are installed with the same version, the command in line 59  `pip install ${WORKSPACE}/atacworks_wheel/*` will actually skip the wheel file installation because this requirement is already satisfied.

ntadimeti: good point! Made the change

ohadmo: @ntadimeti  it [doesn't seem](https://gpuci.gpuopenanalytics.com/blue/organizations/jenkins/clara-genomics%2Fgpuci%2Fatac-works%2Fprb%2Fatac-works-cpu-build/detail/atac-works-cpu-build/576/pipeline) this syntax works, you need to pass the package name explicitly.

ntadimeti:
@ohadmo right, my bad. Changed it to atacworks. If everything else looks good, I can keep an eye on the CI and merge the PR once it passes and looks good. 

Any other changes needed in this PR ?

PR #181, opened 2020-07-07
WIP: Remove individual scripts and modularize the pre-processing steps

ntadimeti:


ntadimeti:
This PR is very big just from the nature of the change. 
1) get_intervals.py, bw2h5.py, peak2bw.py files are now modules. All of them are called inside main.py.
2) Due to the change above, a bunch of new options are exposed through config files and some options that are no longer relevant are removed (files_train, val_files, etc). 
3) Some config options like print_freq, eval_freq, debug, transform, clip_grad etc were considered unimportant and removed after brainstorming with @avantikalal 
4) --sizes_file is renamed as --genome. It can now take in strings "hg19" or "hg38" for picking up the sizes files automatically. Optionally one can also provide a path to a sizes file. 
5) --regions option is added. This is only relevant for denoising, not for training. Using this, one can choose a subset of chromosomes to run denoising on. See help for how it works. 
6) All tests are updated to reflect this API. expected_results folder files are updated to reflect the new output folder structure.

@avantikalal @tijyojwad -- I suggest you to take a look at this PR as soon as possible, as I suspect this to take some time to iterate on and merge

atacworks/dl4atac/utils.py
```diff
@@ -198,7 +198,7 @@ def progbar(*, curr, total, progbar_len, pre_bar_msg, post_bar_msg):
     # sys.stdout.flush()
 
 
-def gather_files_from_cmdline(input, extension=".h5"):
+def gather_files_from_cmdline(input, extension=(".h5", ".bed", ".narrowPeak")):
```
tijyojwad, line 201: Thinking about this, I feel like it would be better if client had to always specify the extension(s)

configs/infer_config.yaml
```diff
@@ -8,44 +8,73 @@
 # license agreement from NVIDIA CORPORATION is strictly prohibited.
 #
 
+### Data Pre-processing parameters
+## Generate intervals files
+intervalsize: 50000
```
tijyojwad, line 13: there's a bit of inconsistency in our naming scheme where some options are `wordword` and some are `word_word`, like `intervalsize` vs `batch_size`. We should make the consistent (and I personally like the latter better)

ntadimeti:
rerun tests

configs/infer_config.yaml
```diff
@@ -8,44 +8,73 @@
...
+wg: False
+# Chromosome sizes to be used
+# You can also provide a file path
```
tijyojwad, line 16: can you add a description of what kind of file this path should be to?

configs/infer_config.yaml
```diff
@@ -8,44 +8,73 @@
...
+# task: regression
+# Similarly, if the task is classification, then model cannot 
+# output denoised tracks. 
+# task: classification
+# The peaks: and tracks: options above are only
+#  relevant if task is "both".
 task: "both"
-print_freq: 50
 bs: 512
+# NUmber of workers to use for multiprocessing
```
ntadimeti, line 62: Fix Number typo

tijyojwad, line 62: `NUmber` -> `Number`

scripts/bw2h5.py
```diff
@@ -70,151 +36,113 @@
 _logger.addHandler(_handler)
 
 
-def parse_args():
-    """Parse command line arguments.
-
-    Return:
-        args : parsed argument object.
+def bw2h5(noisybw, cleanbw, layersbw, cleanpeakbw, batch_size, nolabel,
```
tijyojwad, line 39: how about moving this into `atacworks/io` folder since we call it as an API now?

scripts/get_intervals.py
```diff
@@ -98,129 +69,128 @@ def get_tiling_intervals(sizes, intervalsize, shift=None):
     return intervals.loc[:, ('chrom', 'start', 'end')]
 
 
-def parse_args():
-    """Parse command line arguments.
+def get_intervals(sizesfile, intervalsize, out_dir, wg=None, val=None,
```
tijyojwad, line 72: similarly, how about moving this to `atacworks/dl4atac/utils`?

scripts/main.py
```diff
@@ -321,11 +323,12 @@ def main():
 
     args = parse_args(root_dir)
 
-    # Set log level
-    if args.debug:
-        _handler.setLevel(logging.DEBUG)
-        _logger.setLevel(logging.DEBUG)
+    if args.genome == "hg19":
```
tijyojwad, line 326: for this one, I think this one could make it more extensible - 
```
genomes = {
    "hg19" : os.path.join(root_dir, "reference", "hg19.chrom.sizes"),
    "hg38" : os.path.join(root_dir, "reference", "hg38.chrom.sizes"),
}

if args.genome in genomes:
    return genomes[args.genome]
else:
    # error handling code
```

scripts/main.py
```diff
@@ -345,18 +348,64 @@ def main():
...
         torch.cuda.manual_seed(args.seed)
+
     # train & resume
     ##########################################################################
     if args.mode == "train":
-        args.files_train = gather_files_from_cmdline(args.files_train)
-        args.val_files = gather_files_from_cmdline(args.val_files)
-        _logger.debug("Training data:   " + "\n".join(args.files_train))
+
+        args.cleanpeakfile = gather_files_from_cmdline(args.cleanpeakfile)
```
tijyojwad, line 356: would be good to specify the extension for `cleanpeakfile` just to be explicit about the type

scripts/main.py
```diff
@@ -345,18 +348,64 @@ def main():
...
+        args.noisybw = gather_files_from_cmdline(args.noisybw, extension=".bw")
+        args.cleanbw = gather_files_from_cmdline(args.cleanbw, extension=".bw")
+
+        # We have to make sure there is a 1-1 correspondence between files.
+        assert len(args.cleanpeakfile) == len(args.noisybw)
+        assert len(args.cleanbw) == len(args.noisybw)
+
+        train_files = []
```
tijyojwad, line 364: since `train_files` and `val_files` are generated from respective clean/noisy tracks, is there a way to run training/validation from pre generated .h5 files?

scripts/main.py
```diff
@@ -381,27 +430,55 @@ def main():
...
     ##########################################################################
     if args.mode == "denoise" or args.mode == "eval":
-        files = args.input_files
-        files = gather_files_from_cmdline(files)
+
+        if args.mode == "eval":
+            # Read in the narrowPeak or BED files for clean data peak
+            # labels, convert them to bigwig
+            out_path = os.path.join(args.exp_dir, "bigwig_peakfiles")
+            cleanpeakbw = peak2bw(args.cleanpeakfile, args.genome,
```
tijyojwad, line 438: Is this variable is used in the denoise/eval conditional?

scripts/peak2bw.py
```diff
@@ -49,57 +30,36 @@
 _logger.addHandler(_handler)
 
 
-def parse_args():
-    """Parse command line arguments.
+def peak2bw(input_file, sizesfile, out_dir):
```
tijyojwad, line 33: same, how about moving this over to `atacworks/io`?

setup.py
```diff
@@ -93,6 +93,9 @@ def get_installation_requirments(file_path):
           ('configs', ['configs/infer_config.yaml',
                        'configs/train_config.yaml',
                        'configs/model_structure.yaml']),
-          ('reference', ['reference/hg19.auto.sizes'])],
+          ('reference', ['data/reference/hg19.chrom.sizes',
```
tijyojwad, line 96: does this support wildcards like `data/reference/hg19.*`?

tests/end-to-end/infer.sh
```diff
@@ -15,47 +15,60 @@ echo "Run inference on test set with default peak calling setting..."
...
-    --out_home $out_dir --label inference \
+    --out_home $out_dir --exp_name inference \
     --weights_path $expected_results_dir/model_latest/model_best.pth.tar \
     --num_workers 0 \
-    --input_files $out_dir/no_label.h5 \
-    --intervals_file $out_dir/result.holdout_intervals.bed \
-    --sizes_file $ref_dir/hg19.auto.sizes \
+    --noisybw $data_dir/HSC.5M.chr123.10mb.coverage.bw  \
+    --wg --intervalsize 24000 \
+    --genome $data_dir/example.sizes \
```
tijyojwad, line 23: what do you think of it should be `genome_size` as the name?

scripts/bw2h5.py
ntadimeti: I'm vary of moving it into atacworks/io right now. These "modules" still create files and the module returns a path to the output file instead of an object. I feel like it should qualify as a proper module to graduate into atacworks/io. 

I wanted to return output files instead of objects for 2 reasons: 
1) It's a minimal change, that still let's us test the intermediate files through end-to-end testing.
2) Keeping deadline in mind, this is was the fastest way with least potential errors.

Thoughts?

scripts/main.py
ntadimeti: Right now, API does not allow for that. I felt like it would cause a lot of confusion for users to have a lot of input flexibility. If there is a real cost of time/money involved, we can reconsider this in the future. 
 WDYT ?

scripts/peak2bw.py
ntadimeti: See explanation for bw2h5

tests/end-to-end/infer.sh
ntadimeti: It's really a file that contains sizes. To name it genome_file will be confusing because users can provide just a string "hg19" as well. So, left it as genome for now

scripts/main.py
ntadimeti: It takes in both narrowPeak and BED types. Found it hard to encode that into the name.

atacworks/dl4atac/utils.py
ntadimeti: I thought it would be good to have a list of expected formats, instead of having to always specify it ?

tijyojwad: from an API perspective specifying which extensions the files are to be gathered for is better. I would make that a required argument

scripts/main.py
tijyojwad: can't it be with `extension=("ex1", "ext2")`?

tests/end-to-end/infer.sh
tijyojwad: Yeah, I don't want the name to be `genome_file`, I was suggesting `genome_size` :). And size could be either a file, or an alias (like "hg19")

scripts/bw2h5.py
tijyojwad: After discussing offline, decided to move into an `atacworks` module to enable better testing and prepping for future improvements as an API

tests/end-to-end/infer.sh
avantikalal: genome_size is confusing because users may think it means the size of the genome (3 billion bp).

configs/infer_config.yaml
```diff
@@ -8,44 +8,69 @@
...
 #
 
+### Data Pre-processing parameters
+## Generate intervals files
+intervalsize: 50000
+wg: False
+# Chromosome sizes to be used
+# You can also provide a file path
+genome: "hg19"
+# Ratio of non-peak to peak intervals in training
```
avantikalal, line 18: Why is this option part of `infer_config`? Seems relevant only for training.

scripts/bw2h5.py
```diff
@@ -10,43 +10,10 @@
...
-            --intervals validation_intervals.bed \
-            --out_dir ./ --prefix validation_data \
-            --cleanbw clean.bw --cleanpeakbw clean.narrowPeak.bw
-    Inference:
-        python bw2h5.py --noisybw noisy.bw --intervals test_intervals.bed \
-            --out_dir ./ --prefix test_data \
-            --nolabel
-
-"""
+r"""Read data from bigWig files and generate batch data for training."""
```
avantikalal, line 13: "generate batch data" > "generate encoded data"

scripts/bw2h5.py
```diff
@@ -70,151 +36,113 @@
...
+
+    Args:
+        noisybw: BigWig file containing noisy data.
+        cleanbw: BigWig file containing clean data.
+        layersbw: BigWig file containing layers data.
+        cleanpeakbw: BigWig file containing clean peaks data to be used as
+        labels for training.
+        batch_size: Number of lines to read at a time, since all of the data
+        does not fit in memory.
+        nolabel: No peak labels to encode into h5. Used for generating
```
avantikalal, line 51: Do we need this option? Can we assume that if label files are not provided, we should encode the noisy data alone?

scripts/cmd_args.py
```diff
@@ -83,13 +84,49 @@ def add_common_options(parser):
         parser : After adding the arguments.
 
     """
+    # Pre-processing 1 : Get intervals
```
avantikalal, line 87: This seems to cover bw2h5 as well as get intervals

scripts/cmd_args.py
```diff
@@ -83,13 +84,49 @@ def add_common_options(parser):
...
+                        (low coverage/low quality) ATAC-seq signal',
+               required=True)
+    parser.add('--cleanbw', type=type_or_none_fn(str),
+               help='Path to bigwig file containing clean \
+                        (high-coverage/high-quality) ATAC-seq signal.\
+                            Not used with --nolabel.')
+    parser.add('--cleanpeakfile', type=type_or_none_fn(str),
+               help='Path to narrowPeak or BED file containing peak calls '
+                    'from \
+                        MACS2.')
```
avantikalal, line 100: "from MACS2 on the clean (high-coverage/high-quality) ATAC-seq signal.

scripts/cmd_args.py
```diff
@@ -105,9 +142,6 @@ def add_common_options(parser):
                    on either side of each interval. Use the same --pad \
                    value that was supplied to bw2h5.py when creating \
                    the h5 files for training and validation.")
-    parser.add('--transform', required=True, type=str, choices=['log', 'None'],
-               help='transformation to apply to\
-                           coverage tracks before training')
     parser.add('--layers', required=True, type=type_or_none_fn(str),
```
avantikalal, line 145: Is it necessary to supply this again? We can use the names that were supplied with the `layersbw` option.

configs/infer_config.yaml
```diff
@@ -8,44 +8,69 @@
...
+#intervals that are equally spaced with intervalsize
+#Look at this file for reference:
+#https://github.com/clara-parabricks/AtacWorks/blob/master/tests/expected_results/end-to-end/result.holdout_intervals.bed
+regions: "None"
+
+## Input files
+noisybw: "None"
+cleanbw: "None"
+layersbw: "None"
+# Batch size to read the bigiwig files, because
```
avantikalal, line 38: Some explanations of arguments are given in the config file while others are given in the help options in `cmd_args.py`. For consistency, should we keep all explanations in the help options alone and ask the user to use `-h` to understand the parameters?

scripts/cmd_args.py
```diff
@@ -199,43 +228,51 @@ def add_inference_options(parser):
...
+    parser.add('--wg', action='store_true',
+               help='Set this flag to produce one set of intervals for\
+                     whole genome. If you would like to run denoising \
+                     on a subset of chromosome instead, take a look at \
+                     the --regions option instead.')
+    parser.add('--regions', required=True, type=type_or_none_fn(str),
+               help='atacworks denoising is done on whole genome by'
+                    'default. You can optionally specify a list of \
+                     chromosomes separated by comma and no spaces \
+                     like [chr1,chr2]. You can aslo provide list of \
```
avantikalal, line 241: also

scripts/cmd_args.py
```diff
@@ -199,43 +228,51 @@ def add_inference_options(parser):
...
+                     [chr1:0-1000,chr2,chr3:0-500]. Please note \
+                     NO SPACES. You can also provide a BED file that \
+                     contains equally spaced chromosome intervals. \
+                     See documentation for creating your own regions \
+                     file.')
     parser.add('--peaks', action='store_true',
-               help='Output denosied peaks from atacworks. If --task is regression, \
-                       model only outputs denoised tracks and \
-                       this option becomes irrelevant.')
+               help='Set this flag to output denosied peaks from atacworks. '
```
avantikalal, line 249: denoised

scripts/cmd_args.py
```diff
@@ -199,43 +228,51 @@ def add_inference_options(parser):
...
                        for regression outputs')
-    parser.add('--cla_rounding', required=True, type=int,
-               help='number of decimal digits to round values \
-                       for classification outputs')
     parser.add('--batches_per_worker', required=True, type=int,
                help='number of batches to run per worker\
                                during multiprocessing')
     parser.add('--gen_bigwig', action='store_true',
-               help='save the inference output to bigiwig\
+               help='Save the inference output to bigiwig\
```
avantikalal, line 274: bigwig

scripts/get_intervals.py
```diff
@@ -98,129 +69,129 @@ def get_tiling_intervals(sizes, intervalsize, shift=None):
...
     # Read chromosome sizes
-    sizes = read_sizes(args.sizes)
+    sizes = read_sizes(sizesfile)
 
+    # Create the output dir
+    if not os.path.exists(out_dir):
+        os.makedirs(out_dir)
     # Generate intervals
-    if args.wg:
+    if wg:
```
avantikalal, line 99: Do we need the --wg option? Could we just assume that if chromosomes are unspecified then we should use the whole genome?

scripts/main.py
```diff
@@ -381,27 +430,55 @@ def main():
...
+
         for x in range(len(files)):
             infile = files[x]
             args.input_files = [infile]
             if args.mode == "denoise":
-                _logger.debug("Inference data: ", args.input_files)
+                _logger.debug("Inference data: ", infile)
 
                 # Check that intervals, sizes and h5 file are all compatible.
                 _logger.info('Checkng input files for compatibility')
```
avantikalal, line 470: Checking

scripts/get_intervals.py
```diff
@@ -98,129 +69,129 @@ def get_tiling_intervals(sizes, intervalsize, shift=None):
...
+                chrom = region
+                chrom_range = None
+                # If regions are specified with intervals like chr1:0-50
+                # Then split the region into chrom and it's range.
+                if region.find(":") != -1:
+                    chrom, chrom_range = region.split(":")
+                chrom_sizes = sizes[sizes['chrom'] == chrom]
+                intervals = get_tiling_intervals(chrom_sizes, intervalsize)
+
+                if chrom_range:
```
avantikalal, line 184: This behavior wouldn't denoise the regions asked for by the user. If, for example, intervalsize is 1000, and the user specifies "chr1:1500-2500"
- the intervals produced by L182 would be:
chr1 0 1000
chr1 1000 2000
chr1 2000 3000
- and none of these intervals would be selected by L186-189.

We should retain any intervals that intersect at all with the specified region. 
Should also explain this behavior in the help / docs.

ntadimeti:
Closing this PR since it has become too big to track the changes. Will open several smaller PRs which contain subset of these changes for easy review and merge

PR #182, opened 2020-07-11
Add unit testing for bedio.py and h5io.py

ntadimeti:


tests/io/test_bedio.py
```diff
@@ -0,0 +1,100 @@
+#
+# Copyright (c) 2019, NVIDIA CORPORATION.  All rights reserved.
```
tijyojwad, line 2: 2020

tests/io/test_h5io.py
```diff
@@ -0,0 +1,54 @@
+#
+# Copyright (c) 2019, NVIDIA CORPORATION.  All rights reserved.
```
tijyojwad, line 2: 2020

tests/io/test_bedio.py
```diff
@@ -0,0 +1,100 @@
...
+#
+# NVIDIA CORPORATION and its licensors retain all intellectual property
+# and proprietary rights in and to this software, related documentation
+# and any modifications thereto.  Any use, reproduction, disclosure or
+# distribution of this software and related documentation without an express
+# license agreement from NVIDIA CORPORATION is strictly prohibited.
+#
+
+from atacworks.io import bedio
```
tijyojwad, line 11: update the ordering of imports to be 

```
standard modules

3rd party modules

own package modules
```

e.g.
```
import os

import pandas as pd
import pytest

from atacworks.io import bedio
```

tests/io/test_h5io.py
```diff
@@ -0,0 +1,54 @@
...
+#
+# NVIDIA CORPORATION and its licensors retain all intellectual property
+# and proprietary rights in and to this software, related documentation
+# and any modifications thereto.  Any use, reproduction, disclosure or
+# distribution of this software and related documentation without an express
+# license agreement from NVIDIA CORPORATION is strictly prohibited.
+#
+
+from atacworks.io import h5io
```
tijyojwad, line 11: same import reordering comment as the previous one

tests/io/test_bedio.py
```diff
@@ -0,0 +1,100 @@
...
+
+import pandas as pd
+import pytest
+import os
+
+
+@pytest.mark.cpu
+def test_read_intervals(tmpdir):
+    """ Test read intervals."""
```
tijyojwad, line 20: general comment for all the tests - it would be good to have docstrings which explain a brief of what the test is doing

e.g.
```
This test checks for correctness of read_intervals API by writing intervals to a bed file and then reading in the same file and checking for equivalance.
```

tests/io/test_h5io.py
```diff
@@ -0,0 +1,54 @@
...
+    with h5py.File(h5file, 'r') as f:
+        for key in input_dict.keys():
+            data = np.array(f[key])
+            assert data.all() == input_dict[key].all()
+
+    # Append to existing dataset and test.
+    input_dict = {"features": np.array([0, 1, 2, 3]),
+                  "scores": np.array([1, 2, 1, 1])}
+    new_dict = {"features": np.array([0, 1, 2, 3, 4, 5, 6])}
+    input_dict.update(new_dict)
```
tijyojwad, line 35: How about this to update the `features entry`
```
input_dict["features"] = np.array([0, 1, 2, 3, 4, 5, 6])
```

tests/io/test_h5io.py
```diff
@@ -0,0 +1,54 @@
...
+    """Test h5 to array."""
+
+    input_dict = {"features": np.array([0, 1, 2, 3]),
+                  "scores": np.array([1, 2, 1, 1])}
+    h5file = os.path.join(tmpdir, "h5file.h5")
+    with h5py.File(h5file) as f:
+        for key in input_dict.keys():
+            f.create_dataset(key, data=input_dict[key])
+    data = h5io.h5_to_array(h5file, "features", None)
+    assert data.all() == input_dict["features"].flatten().all()
```
tijyojwad, line 54: neat! can you update this to a loop that goes over all the keys in the original `input_dict` and runs the test for each?

tests/io/test_bedio.py
ntadimeti: done.

tests/io/test_h5io.py
ntadimeti: done.

tests/io/test_bedio.py
ntadimeti: done.

tests/io/test_h5io.py
ntadimeti: done.

tests/io/test_bedio.py
ntadimeti: done.

tests/io/test_h5io.py
ntadimeti: You think it's better to manually update the input_dict, instead of updating using new_dict ? It depends on whether we trust the `.update` or not. I think both are fine, if you have a strong preference, I'll change it.

tests/io/test_h5io.py
ntadimeti: done.

tests/io/test_h5io.py
tijyojwad: I was mostly wondering if there's a reason to use `update` instead of overwriting? Okay either way

tests/io/test_bedio.py
```diff
@@ -0,0 +1,115 @@
...
+#
+# NVIDIA CORPORATION and its licensors retain all intellectual property
+# and proprietary rights in and to this software, related documentation
+# and any modifications thereto.  Any use, reproduction, disclosure or
+# distribution of this software and related documentation without an express
+# license agreement from NVIDIA CORPORATION is strictly prohibited.
+#
+
+import os
+import pandas as pd
```
tijyojwad, line 12: A new line between the different categories helps distinguish the various imports

tests/io/test_h5io.py
ntadimeti: So, I send in a "new_dict" dictionary to verify that the API is actually appending to the existing dataset instead of overwriting it. So, I update the `input_dict` to reflect that.

tests/io/test_bedio.py
ntadimeti: done

tests/io/test_h5io.py
ntadimeti: done!

PR #185, opened 2020-07-15
Add unit testing for bigwigio.py and bedgraphio.py

rmovva:


tests/io/test_bedgraphio.py
```diff
@@ -0,0 +1,140 @@
...
+                                    [0.1, 0.1, 0.0]]}
+    intervals_df = pd.DataFrame(intervals_df_dict)
+    output_df = bedgraphio.intervals_to_bg(intervals_df, resolution=None)
+    output_df.reset_index(drop=True, inplace=True)
+    expected_df_dict = {'chrom': ['chr1'] * 4 + ['chr2'],
+                        'start': [1, 3, 4, 6, 2],
+                        'end': [3, 4, 6, 7, 4],
+                        'score': [0.5, 0.1, 0.5, 0.1, 0.1]}
+    expected_df = pd.DataFrame(expected_df_dict)
+    pd.testing.assert_frame_equal(expected_df, output_df, check_dtype=False)
```
ntadimeti, line 84: Is there a reason why `equals` API was not used ?

rmovva: `equals` is too stringent, e.g. requiring dtypes to be equal, so it's just more convenient to use `pandas` dataframe-specific function which is specifically designed for this purpose

rmovva: if it's poor style to use `assert_frame_equal()` here and `df.equals()` in the first two tests, I could change all to `assert_frame_equal()`?

tijyojwad: Agreed, I think we can switch to `assert_frame_equal` for everything - unless @ntadimeti you feel we need the stringent checking?

tijyojwad:
overall changes look good to me. once @ntadimeti approves it can be merged

tests/io/test_bedgraphio.py
ntadimeti: Discussed this with Raj. I would ideally like to keep it stringent, since we expect certain dtypes in the code. If for some reason that's hard to test, I'll look into it. I'm taking over this PR from Raj, as he has other things to get to.

PR #187, opened 2020-07-17
Remove unncessary args

ntadimeti:


scripts/worker.py
```diff
@@ -215,42 +216,40 @@ def train_worker(gpu, ngpu_per_node, args, timers=None):
         train(rank=rank, gpu=gpu, task=args.task, model=model,
               train_loader=train_loader,
               loss_func=loss_func, optimizer=optimizer, epoch=epoch,
-              epochs=args.epochs, clip_grad=args.clip_grad,
```
avantikalal, line 218: Why not remove the clip_grad argument from this function altogether?

ntadimeti: good point, can do that.

ntadimeti: done

PR #189, opened 2020-07-19
Add regions flag

ntadimeti:
Adds --regions flag for denoising a subset of chromosomes in genome file.


scripts/cmd_args.py
```diff
@@ -236,6 +236,17 @@ def add_inference_options(parser):
...
                is_config_file=True, help='config file path')
+    parser.add('--regions', required=True, type=type_or_none_fn(str),
+               help='atacworks denoising is done on whole genome by \
+                     default. You can optionally specify a list of \
+                     chromosomes separated by comma and no spaces \
+                     like [chr1,chr2]. You can also provide list of \
+                     region indices with each chromosome like \
+                     [chr1:0-1000,chr2,chr3:0-500]. Please note \
+                     NO SPACES. You can also provide a BED file that \
+                     contains equally spaced chromosome intervals. \
```
avantikalal, line 247: change to: "a BED file that contains genomic intervals of length equal to --interval_size. The BED file should have three columns (chromosome, start position, and end position) and no header."

atacworks/dl4atac/utils.py
```diff
@@ -440,6 +443,12 @@ def _get_tiling_intervals(sizes, intervalsize):
         chrom = sizes.iloc[i, 0]
         chrend = sizes.iloc[i, 1]
         starts = range(0, chrend - (intervalsize + 1), intervalsize)
+        if chrom_range is not None:
```
avantikalal, line 446: Let's not embed this within `_get_tiling_intervals`. It serves a different purpose and seems completely disconnected from the original function. I'd suggest making this a separate function that gets called if regions are supplied.

scripts/cmd_args.py
```diff
@@ -236,6 +236,17 @@ def add_inference_options(parser):
     add_common_options(parser)
     parser.add('--config', required=False,
                is_config_file=True, help='config file path')
+    parser.add('--regions', required=True, type=type_or_none_fn(str),
```
avantikalal, line 239: Should we specify that the length of the region needs to be a multiple of `--interval_size`?

scripts/cmd_args.py
ntadimeti: done.

atacworks/dl4atac/utils.py
ntadimeti: done.

scripts/cmd_args.py
ntadimeti: good point. Done

atacworks/dl4atac/utils.py
```diff
@@ -525,6 +554,41 @@ def get_intervals(sizesfile, intervalsize, out_dir, val=None,
...
+            for region in regions:
+                # If regions are specified with intervals like chr1:0-50
+                # Then split the region into chrom and it's range.
+                if region.find(":") != -1:
+                    chrom, chrom_range = region.split(":")
+                    chrom_range = chrom_range.split("-")
+                    chrom_range = [int(value) for value in chrom_range]
+                    chrom_sizes = sizes[sizes['chrom'] == chrom]
+                    intervals = _tile_region_intervals(
+                        chrom_sizes.iloc[0,0],
```
avantikalal, line 573: Why do we need to use the sizes file here? The name of the chromosome is already supplied by the user as part of `--regions` and is stored in the `chrom` variable.

ntadimeti: Good catch, left over from prev changes. Removed it.

atacworks/dl4atac/utils.py
```diff
@@ -525,6 +555,40 @@ def get_intervals(sizesfile, intervalsize, out_dir, val=None,
...
+            final_intervals = pd.DataFrame()
+            regions = regions.strip("[]").split(",")
+            for region in regions:
+                # If regions are specified with intervals like chr1:0-50
+                # Then split the region into chrom and it's range.
+                if region.find(":") != -1:
+                    chrom, chrom_range = region.split(":")
+                    chrom_range = chrom_range.split("-")
+                    chrom_range = [int(value) for value in chrom_range]
+                    intervals = _tile_region_intervals(
```
tijyojwad, line 572: optional: it seems to me that this function can be used in both cases. for the full chromosome case, isn't the range simply "0-chr_length"?

ntadimeti: Done!

PR #191, opened 2020-07-20
Add otuput resolution

ntadimeti:


scripts/cmd_args.py
```diff
@@ -239,6 +239,16 @@ def add_inference_options(parser):
                help='prefix for the inference result files.')
     parser.add('--deletebg', action='store_true',
                help='delete output bedGraph file')
+    parser.add('--out_resolution', required=False,
+               type=type_or_none_fn(int),
+               help='resolution of output files. default 1bp. \
+                     Atacworks always denoises at 1 base pair \
+                     resolution. If out_resolution is 5, then \
+                     the coverage values for every 5 rows is \
```
avantikalal, line 247: every 5 rows -> every 5 base pairs

scripts/cmd_args.py
```diff
@@ -239,6 +239,16 @@ def add_inference_options(parser):
...
+                     averaged and that becomes the new coverage \
+                     value for every 5 rows. Keep in mind that the \
```
avantikalal, line 249: Delete "and that becomes the new coverage value for every 5 rows"

scripts/cmd_args.py
ntadimeti: done.

scripts/cmd_args.py
ntadimeti: done.

PR #200, opened 2020-07-28
SA review

ntadimeti:


README.md
```diff
@@ -47,9 +46,10 @@ git clone --recursive https://github.com/clara-genomics/AtacWorks.git
 * Install pip dependencies
 
     ```
-    pip install -r requirements.txt
+    cd AtacWorks && pip install -r requirements.txt
     ```
-* Optional -- Install macs2
+* Optional -- Install macs2.
```
avantikalal, line 51: Change to : Only required if you want to use macs2 subcommands to call peaks based on peak probabilities generated by AtacWorks.

README.md
```diff
@@ -62,13 +62,36 @@ git clone --recursive https://github.com/clara-genomics/AtacWorks.git
     ```
 ### 3. Tests
 
-Run unit tests:
+Run unit tests to verify that installation was successful
```
avantikalal, line 65: Say that this is optional.

README.md
```diff
@@ -62,13 +62,36 @@ git clone --recursive https://github.com/clara-genomics/AtacWorks.git
...
 
     ```
     python -m pytest tests/
     ```
 
+## Workflow
+AtacWorks trains a deep neural network to learn a mapping between noisy (low coverage/low cell count/low quality) ATAC-seq data and matching clean (high coverage, high cell count, and/or high quality) ATAC-seq data. Both noisy and clean data should be from the same cell type or tissue. Once this mapping is learned, the trained model can be applied to improve other noisy ATAC-Seq datasets.
+
+### 1. Training an AtacWorks model
```
avantikalal, line 74: Should we link to the html documentation instead of the notebooks?

ntadimeti: These links don't point to the notebooks, they are pointing to md files (which point to notebooks).

To clarify, if the Readme is read through the webpage, it automatically opens the html page (because sphinx compiles and cross-references it). If the Readme is read through the github repo, then it opens the md files.

README.md
ntadimeti: I think it's important to run the unit tests to make sure all the dependencies are installed properly

README.md
ntadimeti: done.

PR #215, opened 2020-08-28
fix evaluation

ntadimeti:
fixed atacworks eval, added end-to-end test


scripts/main.py
```diff
@@ -480,6 +479,10 @@ def main():
             # for training.
             args.noisybw = gather_files_from_cmdline(args.noisybw,
                                                      extension=".bw")
+            if args.mode == "eval":
+                args.cleanbw = gather_files_from_cmdline(args.cleanbw,
```
tijyojwad, line 483: there's already a condition for `mode == eval` a few lines above - can this be placed there?

scripts/main.py
```diff
@@ -529,25 +531,27 @@ def main():
...
-                            "batches_per_worker": args.batches_per_worker,
-                            "gen_bigwig": args.gen_bigwig,
-                            "sizes_file": args.genome,
-                            "res_queue": res_queue, "prefix": prefix,
-                            "deletebg": args.deletebg,
-                            "out_resolution": args.out_resolution}
-            write_proc = mp.Process(target=writer, kwargs=keyword_args)
-            write_proc.start()
+
+            if args.mode == "denoise":
```
tijyojwad, line 535: what does this condition catch for?

ntadimeti: Writer is only to be called for denoise mode. It's not relevant for eval, there's nothing being written to the `res_queue`. Not sure why we pass `res_queue` to eval_worker at all.

scripts/main.py
ntadimeti: good catch, changed it.

scripts/main.py
```diff
@@ -467,6 +466,9 @@ def main():
                 out_path = os.path.join(args.exp_dir, "bigwig_peakfiles")
                 cleanpeakbw = peak2bw(args.cleanpeakfile, args.genome,
                                       out_path)
+                args.cleanbw = gather_files_from_cmdline(
+                    args.cleanbw,
+                    extension=".bw")
 
             # Generate training, validation, holdout intervals files
```
avantikalal, line 473: Let's change this comment to avoid confusion - we aren't generating training and val files here

ntadimeti: done

PR #245, opened 2021-08-24
Removing hard-set versions from requirements.txt to be more compatible with RAPIDS

cjnolet:
I'm currently updating the `rapids-single-cell-examples` repository to use RAPIDS 21.08 but the firm versions in the atacworks python package are causing numerous conflicts with the rest of the packages in the ecosystem. I propose we remove the hard-set versions since, as far as I can tell, nothing in this repository requires them. If we do that there are version issues over time, we should consider setting bounds with ranges instead of locking to an exact version.


edawson:
I agree these shouldn't be fixed to specific versions and should eventually be a range, but can we use a ">=" semantic instead as a start?

Very much looking forward to seeing this get repaired and released!

cjnolet:
@edawson I made the requested change and verified that the environment still works after building.

edawson (approved):
Versions updated to `>=` syntax as requested and @cjnolet states environment is still happy. Looks good to me :)

